PR #5, opened 2017-11-17
Version 1.0 PowerShell Module with Pester Tests and Markdown Help

sreedatjc:
Version 1.0 of the JumpCloud PowerShell module. Tested against PowerShell Core 6.0 and PowerShell 5. Pester tests included.


scottd3v:
Updated the groups to check for case sensitivity. Updated tests to be easier to use.

PR #26, opened 2018-07-06
Update verify_serviceaccount.sh

kyle-randolph:
Added a check to verify OS version


scottd3v:
Tested. The script works as advertised.

PR #33, opened 2018-08-01
Create hide_mac_app.sh

kyle-randolph:
This short script hides the Mac App from displaying for JumpCloud managed users. This must be run as root on systems where you would like to hide the Mac App.

### DIFF
--- a/scripts/macos/hide_mac_app.sh
+++ b/scripts/macos/hide_mac_app.sh
@@ -1,0 +1,8 @@
+defaults write /Library/LaunchAgents/com.jumpcloud.jcagent-tray Disabled -bool true
+mkdir -p /opt/jc_user_ro
+touch /opt/jc_user_ro/dont_show_tray_app
+PERMISSIONS="$(ls -l /opt/jc_user_ro/dont_show_tray_app | cut -c -10)"
+if [ "$PERMISSIONS" != "-rw-r--r--" ]; then
+        echo "Permissions not right for file:  $PERMISSIONS"
+        exit 1
+fi
--- a/scripts/macos/hide_mac_app.sh
+++ b/scripts/macos/hide_mac_app.sh
@@ -1,8 +1,8 @@
+#!/bin/bash
+if [ $UID != 0 ]; then
+  (>&2 echo "Error:  $0 must be run as root")
+  exit 1
+fi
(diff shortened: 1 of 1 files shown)


scripts/macos/hide_mac_app.sh
```diff
@@ -0,0 +1,8 @@
+defaults write /Library/LaunchAgents/com.jumpcloud.jcagent-tray Disabled -bool true
+mkdir -p /opt/jc_user_ro
+touch /opt/jc_user_ro/dont_show_tray_app
+PERMISSIONS="$(ls -l /opt/jc_user_ro/dont_show_tray_app | cut -c -10)"
+if [ "$PERMISSIONS" != "-rw-r--r--" ]; then
+        echo "Permissions not right for file:  $PERMISSIONS"
```
jc-rob: indentations should be two spaces, we use the google shell style guide as a rule https://google.github.io/styleguide/shell.xml.

scripts/macos/hide_mac_app.sh
```diff
@@ -0,0 +1,8 @@
+defaults write /Library/LaunchAgents/com.jumpcloud.jcagent-tray Disabled -bool true
+mkdir -p /opt/jc_user_ro
+touch /opt/jc_user_ro/dont_show_tray_app
+PERMISSIONS="$(ls -l /opt/jc_user_ro/dont_show_tray_app | cut -c -10)"
+if [ "$PERMISSIONS" != "-rw-r--r--" ]; then
```
jc-rob: are these permissions critical for this script to accomplish its purpose? If so, instead of just bailing if they aren't correct, why not set the perms if wrong?

scripts/macos/hide_mac_app.sh
```diff
@@ -0,0 +1,8 @@
+defaults write /Library/LaunchAgents/com.jumpcloud.jcagent-tray Disabled -bool true
+mkdir -p /opt/jc_user_ro
+touch /opt/jc_user_ro/dont_show_tray_app
+PERMISSIONS="$(ls -l /opt/jc_user_ro/dont_show_tray_app | cut -c -10)"
```
jc-rob: quotes around a stored command isn't necessary

tristanmillerjumpcloud:
Anything else, @jc-rob ?

PR #35, opened 2018-08-13
script to revert sshd to OR behavior

jc-rob:


scripts/linux/mod_sshd_config.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+if [[ $UID != 0 ]]; then
```
pnawalramka: Enclose variable `$UID` in quotes.

scripts/linux/mod_sshd_config.sh
```diff
@@ -0,0 +1,39 @@
...
+  PasswordAuthentication yes
+  PubkeyAuthentication yes
+  AuthenticationMethods password publickey
+# GlobalExceptionsEnd
+EOF
+
+check=$(systemctl --version >/dev/null 2>&1)
+
+if [[ "${check}" = 0 ]]; then
+  restart=$(systemctl restart sshd)
```
pnawalramka: Maybe take out "sshd" as a variable declaration.

PR #37, opened 2018-08-15
Rename Mac Script v2

scottd3v:
script to update mac username


scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -1,13 +1,21 @@
...
+
+ If the user recieving the name change is signed in they will be signed out. 
+
+ Example usage: 
+
+ sudo sh RenameMacUserNameAndHomeDirectory.sh cat dog
+
+ Above example would rename account cat to dog
+
+ NOTE: SCRIPT MUST BE RUN AS SUDO!
```
rmulley: `RUN AS ROOT`

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
 oldUser=$1
 newUser=$2
 
+#Verify valid username
 if [[ -z "${newUser}" ]]
 then
-	abort "New user name must not be empty!"
+	echo "New user name must not be empty!"
```
rmulley: Indentation is off

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
+    exit 1
+fi
+
+#Test to ensure account update is needed
+if [ "$oldUser" = "$newUser" ]; then
+    echo "Current username = $oldUser -eq new username $newUser"
```
rmulley: It may be clearer to remove this line, it's a bit confusing the way it reads.

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -1,13 +1,21 @@
 #!/bin/bash
```
rmulley: We should add the `e` and `u` flags here? For example: `#!/bin/bash -eu`

That would cause the script to exit on any command failures and complain about unused variables. It helps make Bash act more like a "real" language.

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
rmulley: Do we just want to allow this and let it fail later? I'm sure MacOS doesn't allow empty usernames so it should error out anyway?

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
+
+#Test to ensure account update is needed
+if [ "$oldUser" = "$newUser" ]; then
+    echo "Current username = $oldUser -eq new username $newUser"
+    echo "No updates needed"
+    exit 0
+ fi
+
+#Query existing user accounts
+existingUsers=($(dscl . -list /Users | grep -Ev "^_|com.*|root|nobody|daemon|\/" | cut -d, -f1 | sed 's|CN=||g'))
```
rmulley: `readonly` here would be nice to prevent this from being edited accidentally

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
+
+#Logs out user if they are logged in by killing their loginwindow process
+if [[ "${killProcess}" ]]
+then   
+    sudo kill -9 $killProcess
+    echo "$oldUser account was logged in. User logged off to complete username update."
+fi
+
+#Captures current "RealName" this is the displayName
+FullRealName=$(dscl . -read /Users/${oldUser} RealName)
```
rmulley: I'd expect `fullRealName` based on the other variable names

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
+    exit 1
+fi
+
+#Checks if user is logged in 
+killProcess=$(ps -Ajc | grep $oldUser | grep loginwindow | awk '{print $2}')
+
+#Logs out user if they are logged in by killing their loginwindow process
+if [[ "${killProcess}" ]]
+then   
+    sudo kill -9 $killProcess
```
rmulley: If this script needs to be ran as `root` then this `sudo` is unnecessary

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
+#Formats "RealName"
+origRealName=$(echo $FullRealName | cut -d' ' -f2-)
+
+#Updates "RealName" to new username (Yes JCAgent will overwrite this after user/system association)
+sudo dscl . -change "/Users/${oldUser}" RealName "${origRealName}" "${newUser}"
+
+err=$?
+if [[ ${err} -ne 0 ]]
+then
+	echo "Could not rename the user's RealName in dscl. - err=${err}"
```
rmulley: Indentation

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
 
-mv /Users/${oldUser} /Users/${newUser}
+#Updates name of home directory to new username
+mv "${origHomeDir}" "/Users/${newUser}"
+
 err=$?
 if [[ ${err} -ne 0 ]]
 then
-	abort "Could not rename the user's home directory in /Users - the user may not be able to login unless you correct dscl to point back to /Users/${oldUser}"
+	echo "Could not rename the user's home directory in /Users - the user may not be able to login unless you correct dscl to point back to /Users/${oldUser}"
```
rmulley: If we hit this point do we just want to correct it for them?

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
-	abort "Could not rename the user's RecordName in dscl - the user should still be able to login, but with user name ${oldUser}, however, their home directory will be pointed to /Users/${newUser}"
+	echo "Could not rename the user's RecordName in dscl - the user should still be able to login, but with user name ${oldUser}, however, their home directory will be pointed to /Users/${newUser}"
+    exit 1
 fi
 
+#Links old home directory to new. Fixes dock breaking issue
+ln -s "/Users/${newUser}" "/Users/${oldUser}"
+
+#Restarts the Menu Bar app to show the updated username in the fast user switcher drop down
+killall -KILL SystemUIServer
```
rmulley: This seems like it could kill much more than just the JC menu bar app?

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -23,39 +31,119 @@ fi
...
 
-dscl . -change /Users/${oldUser} NFSHomeDirectory /Users/${oldUser} /Users/${newUser}
+#Updates NFS home directory
+sudo dscl . -change "/Users/${oldUser}" NFSHomeDirectory "${origHomeDir}" "/Users/${newUser}"
+
 err=$?
 if [ ${err} -ne 0 ]
 then
-	abort "Could not rename the user's home directory pointer, aborting further changes! - err=${err}"
+	echo "Could not rename the user's home directory pointer, aborting further changes! - err=${err}"
```
rmulley: Indentation

jc-rob:
I had a  bunch of pending comments not submitted, mulley beat me to it. Let's get the syntax and style issues corrected and take a second look.

scottd3v:
@jc-rob @rmulley  I've updated formatted and added requested changes. I've tested the rename script on a physical Macbook and it still works.

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -1,61 +1,137 @@
...
+ Example usage: 
+
+ sudo sh RenameMacUserNameAndHomeDirectory.sh cat dog
+
+ Above example would rename account cat to dog
+
+ NOTE: SCRIPT MUST BE RUN AS ROOT!
+'
+
 abort() {
```
dbyington: Since you're no longer using this function you can remove it, or make use of it. Your choice.

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -1,61 +1,137 @@
...
-then
-	abort "New user name must not be empty!"
+#Verify valid username
+if [[ -z "${newUser}" ]]; then
+	echo "New user name must not be empty!"
+	exit 1
+fi
+
+#Test to ensure account update is needed
+if [ "$oldUser" = "$newUser" ]; then
```
dbyington: There are a few instances of bracket inconsistency, `[ condition ]` vs. `[[ condition ]]`. It would be good to pick one and stick with it. The newer syntax, and I believe currently accepted standard is the `[[ condition ]]` version. 
The major difference is that with double brackets quotes are not strictly required, spaces in values get properly evaluated. But I believe we (JC) have been keeping quotes.

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -1,61 +1,137 @@
...
+if [[ ${err} -ne 0 ]]; then
+	echo "Could not rename the user's RecordName in dscl - the user should still be able to login, but with user name ${oldUser}, however, their home directory will be pointed to /Users/${newUser}"
+	exit 1
 fi
 
+#Links old home directory to new. Fixes dock breaking issue
+ln -s "/Users/${newUser}" "/Users/${oldUser}"
+
+#Restarts the Mac menu bar to show the updated username in the fast user switching menu
+killall -KILL SystemUIServer
```
dbyington: I cringe whenever I see a `-KILL` | `-9` signal being used, unless as a last resort. Probably because I've seen many times where that has left to unclean processes. Is there a more _polite_ way to restart the menu bar?

scottd3v: @dbyington 
I updated the brackets and found a nicer way to log off the user not using -kill 9... I'm searching for a way to restart the menu bar "nicer" Do you think `killall SystemUIServer` is better than `killall -KILL SystemUIServer`? I've run both commands locally and don't notice a difference.

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
```diff
@@ -1,61 +1,137 @@
...
+err=$?
+if [[ ${err} -ne 0 ]]; then
+	echo "Could not rename the user's RealName in dscl. - err=${err}"
+	exit 1
+fi
+
+#Captures current NFS home directory
+origHomeDir=$(dscl . -read "/Users/${oldUser}" NFSHomeDirectory | awk '{print $2}' -)
+
+if [[ -z "${origHomeDir}" ]]; then
```
jc-rob: If this fails, we're already in a condition where changes have been made. It would be best that we check any condition involving a need from oldUser evaluates true before setting changes in motion and/or if any of the steps fail, back them out so the system isn't left in an adverse state.

scripts/macos/RenameMacUserNameAndHomeDirectory.sh
dbyington: Since it appears Apple doesn't provide a way to osascript a message to the `SystemUIServer` to reload/restart I'd probably go with a `SIGHUP`. The `HUP` is the typical signal used to tell something to reload or restart. It appears to work nicely with `SystemUIServer`.

dbyington:
You'll probably have to run something like `git rm .DS_Store` to remove those from the branch, and re-push.

scottd3v:
@jc-rob @rmulley updated per nits.

.gitignore
```diff
@@ -1,3 +1,8 @@
 node_modules/
 build/
 PowerShell/.DS_Store
+PowerShell/JumpCloud Module/.DS_Store
+PowerShell/JumpCloud Module/Docs/.DS_Store
+PowerShell/JumpCloud Module/en-Us/.DS_Store
+.vscode/settings.json
+.vscode/settings.json
```
dbyington: Am I seeing double?
Also, I'd probably just put in `.DS_Store` and that'll catch them all.

scottd3v:
Thanks for the help guys! I've reviewed the comments and implemented a number of the recommendations.

PR #50, opened 2018-12-21
Add script to dedup unix guid uid

bross-jc:
This script detects duplicate `unix_uid` and `unix_guid` values on systemusers for an organization and provides an interface to deduplicate these values.


api-utils/JumpCloud_API_Go_Examples/DedupeSystemUsersUnixUidGuid/dedupeSystemUsersUnixUidGuid.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func incrementUnixValuesOnUsers(api jcapi.JCAPI, userList []jcapi.JCUser, field string, startingValue string, dryRun bool) error {
+	fmt.Printf("Incrementing values on field: %s for %d systemusers\n", field, len(userList))
+
+	// Update User Values, leaving the first in the list alone
+	// Field on the proto is a string, so to make things easier, we convert to int
+	// so math is easier
+	intValue, _ := strconv.Atoi(startingValue)
+	for idx := range userList {
```
ppg: Is there a downside to `for _, user := range userList {` and operating on `user` below? You'd make the function argument `[]*jcapi.JCUser` which I think you want to do anyway.

api-utils/JumpCloud_API_Go_Examples/DedupeSystemUsersUnixUidGuid/dedupeSystemUsersUnixUidGuid.go
```diff
@@ -0,0 +1,216 @@
...
+		if field == "unix_uid" {
+			userList[idx].Uid = strconv.Itoa(intValue)
+		} else { // unix_guid
+			userList[idx].Gid = strconv.Itoa(intValue)
+		}
+		if dryRun {
+			fmt.Printf("Updating user: %s field: %s with value: %d\n", userList[idx].Id, field, intValue)
+		} else { // Update User
+			userId, err := api.AddUpdateUser(3, userList[idx])
```
ppg: What is `3` here? Seems like it should be a constant.

api-utils/JumpCloud_API_Go_Examples/DedupeSystemUsersUnixUidGuid/dedupeSystemUsersUnixUidGuid.go
```diff
@@ -0,0 +1,145 @@
...
+	seenUids := make(map[int]bool)
+	seenGuids := make(map[int]bool)
+	dupeUsers := make([]jcapi.JCUser, 0)
+	uidValues := make([]int, 0)
+	guidValues := make([]int, 0)
+
+	dupeUid := false
+	dupeGuid := false
+	for _, user := range userList {
+		uidInt, _ := strconv.Atoi(user.Uid)
```
bbawn: This treats all non-integer values as 0 (and therefore dups of each other). We recently learned that there are around 2k systemusers in prd with `unix_uid = null` (see https://jumpcloud.atlassian.net/browse/SJ-352).  We'll have to think about this and probably discuss with @ppg but my inclination would be to consider `null` values to be valid (e.g. windows-only systemusers don't need a `uid`). In that case, we would want to omit `null` values from our dup sets.

PR #53, opened 2019-01-08
New/set jc user with mfa data

epanipinto-jc:


epanipinto-jc:
accidentally opened - closing

PR #59, opened 2019-01-16
Updated parameter sets consolidated code

scottd3v:


scottd3v:
Bug fix for Remove-JCUser. The parameter set -ByID and -UserID were not working.

PR #61, opened 2019-01-16
Bug fix for posix groups

scottd3v:


scottd3v:
Updated to allow Radius attributes on JC User groups configured as posix groups

PR #65, opened 2019-01-22
Add script to repair failed Windows 64-bit upgrades

jmontgomery-jc:
This script repairs failed upgrades on 64-bit Windows systems by reconfiguring the service to point to the correct executable and starting the repaired service.

### DIFF
--- a/scripts/windows/RepairFailed64BitMigration.bat
+++ b/scripts/windows/RepairFailed64BitMigration.bat
@@ -4,7 +4,6 @@ rem We don't care about the output of this call, we just use it to verify we are
 net session >nul 2>&1
 if %errorlevel% neq 0 (
     echo Error: This script must be run as administrator
-    pause
     exit /B 1
 )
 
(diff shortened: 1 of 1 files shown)


scripts/windows/RepairFailed64BitMigration.bat
```diff
@@ -0,0 +1,29 @@
+rem This script repairs a broken agent install after encountering the 64-bit OS upgrade bug on v0.10.41 
+@echo off
+rem We don't care about the output of this call, we just use it to verify we are administrator
+net session >nul 2>&1
+if %errorlevel% neq 0 (
+    echo Error: This script must be run as administrator
+    pause
```
tristanmillerjumpcloud: Do we want these pause commands?  They might get in the way of running this unattended.  I think Grab might be one of those that can run stuff even with the agent down.

jmontgomery-jc: I can definitely remove them. I added them out of habit to avoid the "prompt pops up and then disappears" thing, but I can definitely see that point

PR #73, opened 2019-02-19
Power shell module

scottd3v:
Updating -enable_user_portal_multifactor $true to update mfa property on user vs mfaData


scottd3v:
verified on MacOS and Windows

PR #93, opened 2019-04-23
Sj 579

mhoneyman-jc:
We often get log snippets in slack requests that are often not what we need to investigate the issues. The agent_log_colector script gathers all OS relevant files related to the agent and zips them for the user to submit to support.


scripts/macos/agent_diagnostics.sh
```diff
@@ -0,0 +1,145 @@
...
+
+# Is zip installed?
+if ! which zip 2> /dev/null; then
+  ZPATH="false"
+else
+  ZPATH=$(which zip)
+fi
+
+function indent() {
+	sed 's/^/		/g'
```
istevenson-JC: I believe this should be ` sed 's/^/\t\t/g'` as you have in the Linux script.

scripts/macos/agent_diagnostics.sh
```diff
@@ -0,0 +1,145 @@
...
+    INVENTORY+=("${i}")
+  done
+
+  # check to see if zip exists.
+  if [[ "$ZPATH" = "false" ]]; then
+    ZIPIT="zip is not installed. please send the following files with your support request:\n${INVENTORY[*]}"
+  else
+    if [[ -f "${ZIPFILE}" ]]; then
+      mv "${ZIPFILE}" ./jc"${STAMP}".bak.zip
+      zip -r "*.crt" -r "${ZIPFILE}" "${JCPATH}" > /dev/null 1
```
istevenson-JC: As with the Linux script, this should be `zip -x "*.crt" -r "${ZIPFILE}" "${JCPATH}" > /dev/null 1`.

scripts/macos/agent_diagnostics.sh
```diff
@@ -0,0 +1,145 @@
...
+  SUDODIR="/etc/sudoers.d"
+  SUDOLIST=( $(ls ${SUDODIR}) )
+  for i in "${SUDOLIST[@]}"; do
+    SUDOERS+=("${i}")
+  done
+}
+
+function jconf() {
+  # Grab the contents of jconf for quick display in the output.log file.
+  JCAGENTCONFIG=( $(sed 's/[{}]//g' "${JCPATH}"/jcagent.conf | tr ',' '\n') )
```
istevenson-JC: Your method of initializing `JCAGENTCONFIG` is different in the Linux script, (i.e. `JCAGENTCONFIG=( $(sed 's/,/\n/g' "${JCPATH}"/jcagent.conf | sed 's/[{}]//g') )`). Is this intentional?

scripts/macos/agent_diagnostics.sh
```diff
@@ -0,0 +1,145 @@
...
+  for i in "${JCAGENTCONFIG[@]}"; do
+    JCONF+=("${i}")
+  done
+}
+
+function info_out() {
+  # Write the output.log file.
+  SERVICEVERSION=$( cat /opt/jc/version.txt )
+  SYSINFO=$( uname -rs )
```
istevenson-JC: Did we want to collect operating system release information for macOS systems as we do for Linux? The `sw_vers` command provides output similar to the following:
```
ProductName:    Mac OS X
ProductVersion: 10.7.4
BuildVersion:   11E53
```

scripts/linux/agent_diagnostics.sh
```diff
@@ -0,0 +1,152 @@
...
+    INVENTORY+=("${i}")
+  done
+
+  # check to see if zip exists.
+  if [[ "$ZPATH" = "false" ]]; then
+    ZIPIT="zip is not installed. please send the following files with your support request:"
+  else
+    if [[ -f "${ZIPFILE}" ]]; then
+      mv "${ZIPFILE}" ./jc"${STAMP}".bak.zip
+      zip -r "*.crt" -r "${ZIPFILE}" "${JCPATH}" > /dev/null 1
```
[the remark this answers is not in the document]
istevenson-JC: Furthermore, we should exclude the `.key` files as we do in the Windows script.

scripts/macos/agent_diagnostics.sh
istevenson-JC: Furthermore, we should exclude the `.key` files as we do in the Windows script.

scripts/windows/agent_diagnostics.ps1
```diff
@@ -0,0 +1,87 @@
...
+        Copy-Item $_.FullName $TEMP -Recurse -Force -Exclude @("*.crt", "*.key")
+}
+    Compress-Archive -Path $TEMP -DestinationPath $WRITEPATH\$ZIPFILE
+    Remove-Item -Path $TEMP -Recurse
+    return $zipjc_out
+}
+
+function ziplog {
+    # Zip the log files.
+    $LOGFILE="jcagent.log"
```
istevenson-JC: Is there a `jcUpdate.log` on Windows that we should grab as well?

scripts/windows/agent_diagnostics.ps1
```diff
@@ -0,0 +1,87 @@
...
+    foreach($l in $JCAGENTCONFIG) {
+        $jconf_out+=("`t$($l -replace '[{}]', '')`n")
+    }
+    return $jconf_out
+}
+
+function info_out {
+    # Write the output.log file.
+    if(Test-Path $OUTPUTFILE) {
+    mv $OUTPUTFILE $WRITEPATH\output$STAMP.log -Force
```
istevenson-JC: This should be indented as it is within the `if` block. Is PowerShell supported by `shellcheck`? Might be useful to run it, if so, or an equivalent to lint our PowerShell scripts.

mhoneyman-jc: fixed indenting and used PSScriptAnalyzer for style and format.

scripts/windows/agent_diagnostics.ps1
mhoneyman-jc: There are multiple log files I missed on the first pass. I added a loop to get all jc* and Jump* files from c:\Windows\Temp.

scripts/macos/agent_diagnostics.sh
mhoneyman-jc: Changed from uname to sw_vers, as the info is better.

istevenson-JC: You could also use both commands, if the combined information is useful.

scripts/macos/agent_diagnostics.sh
mhoneyman-jc: I wrote the linux version first and had problems with the sed switch of , to \n. This is due to a difference between the BSD and GNU versions of sed. The preferred BSD solution seems to be using tr. This is portable to GNU systems, so I'm going to leave this as is and change the linux version to match.

scripts/macos/agent_diagnostics.sh
mhoneyman-jc: I reworked this part to perform the exclusions when generating the inventory, instead of when updating the zip file. This change will be made in the linux script as well.

scripts/macos/agent_diagnostics.sh
mhoneyman-jc: Due to a difference between bsd sed and gnu sed, I used a literal (ctrl-v + tab), originally. I changed this to be more readable with:
sed 's/^/\'$'\t/g'

scripts/macos/agent_diagnostics.sh
mhoneyman-jc: I reworked the ziplog function to test for the presence of the jumpcloud-loginwindow directory for inclusion in the zip archive.

scripts/linux/agent_diagnostics.sh
```diff
@@ -0,0 +1,152 @@
...
+  for i in "${JCAGENTCONFIG[@]}"; do
+    JCONF+=("${i}")
+  done
+}
+
+function info_out() {
+  # Write the output.log file.
+  SERVICEVERSION=$( cat /opt/jc/version.txt )
+  SYSINFO=$( uname -rs )
+  OS=$( grep PRETTY_NAME /etc/os-release | sed 's/\=/:/g' | cut -d':' -f 2 | sed 's/\"//g' )
```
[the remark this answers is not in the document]
mhoneyman-jc: I opted to go with a version of Rob's recommendation:
OS=$(grep PRETTY_NAME /etc/os-release | cut -d\" -f2)

scripts/linux/agent_diagnostics.sh
```diff
@@ -0,0 +1,152 @@
...
+  for i in "${USERLIST[@]}"; do
+    if ! [[ "${i}" == 'root' ]] && ! [[ "${i}" == 'halt' ]] && ! [[ "${i}" == 'restart' ]]; then
+    	USERS+=("${i}")
+    fi
+  done
+}
+
+function sudoers() {
+  # Get a list of the sudoers list.
+  SUDODIR="/etc/sudoers.d"
```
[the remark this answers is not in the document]
mhoneyman-jc: sudo is on the requirements list for all the linux systems we support. all systems tested appear to have /etc/sudoers.d. I have yet to test against centos6, debian8, and amazon, but will verify those before requesting another review.

scripts/linux/agent_diagnostics.sh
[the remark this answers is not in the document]
mhoneyman-jc: I've moved the exclusions up to the collection of the inventory list, instead of trying to do it inline in the zip command, which seems like a cleaner solution.

currently, I'm checking for the presence of zip, and if it doesn't exist, I simply make note it in the output. I am planning to change this to fall back to tar on systems that don't have zip (tar is a requirement for supported linux systems), but for now will hint to install zip or include the files in the inventory list with the support request.

scripts/linux/agent_diagnostics.sh
```diff
@@ -0,0 +1,152 @@
...
+ZIPFILE="./jc${STAMP}.zip"
+declare -a INVENTORY
+
+# Is jcagent installed? If not, exit.
+if [[ ! -d "${JCPATH}" ]]; then
+  echo "JCAGENT IS NOT INSTALLED ON THIS MACHINE."
+  exit 1
+fi
+
+# Is zip installed?
```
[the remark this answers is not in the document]
mhoneyman-jc: Zip is a more consistent option, as the files are a bit easier to deal with for non-linux/unix users. That being said, tar is a requirement for all Jumpcloud supported linux systems, so I'll add a bit to fall back to tar if zip isn't available.

scripts/linux/agent_diagnostics.sh
```diff
@@ -0,0 +1,152 @@
...
+# Some global variables
+JCPATH="/opt/jc"
+JCLOG="/var/log/"
+STAMP=$( date +"%Y%m%d%H%M%S" )
+ZIPFILE="./jc${STAMP}.zip"
+declare -a INVENTORY
+
+# Is jcagent installed? If not, exit.
+if [[ ! -d "${JCPATH}" ]]; then
+  echo "JCAGENT IS NOT INSTALLED ON THIS MACHINE."
```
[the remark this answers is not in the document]
mhoneyman-jc: All caps seemed appropriate at first, as this particular message was intended to stand out as a fatal error. Looking at it now, from the perspective of being as user friendly as possible, it makes sense to  use normal case.

scripts/linux/agent_diagnostics.sh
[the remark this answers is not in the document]
mhoneyman-jc: added a fall back to use tar for linux and mac systems.

scripts/linux/agent_diagnostics.sh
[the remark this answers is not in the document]
mhoneyman-jc: I also added a fallback for centos 6 and rhel 6 systems as the os-release file isn't standard for those systems.

scripts/linux/agent_diagnostics.sh
```diff
@@ -103,9 +117,17 @@ function info_out() {
...
-  OS=$( grep PRETTY_NAME /etc/os-release | sed 's/\=/:/g' | cut -d':' -f 2 | sed 's/\"//g' )
+  if [[ -f /etc/os-release ]]; then
+    OS=$( grep PRETTY_NAME /etc/os-release | cut -d\" -f2 )
+  else
+    OS=$( grep release /etc/redhat-release )
+  fi
   SERVICE="jcagent"
   STATUS=$( service ${SERVICE} status 2> /dev/null )
+  if [[ -z "${STATUS}" ]]; then
+    STATUS=$( echo "PID  PATH" ; ps ax | grep [j]cagent | awk '{print $1" "$5}')
```
istevenson-JC: What is the purpose of writing `[j]cagent`? The result is the same as having `jcagent`.

scripts/linux/agent_diagnostics.sh
```diff
@@ -16,56 +16,70 @@ JCPATH="/opt/jc"
...
 # Is jcagent installed? If not, exit.
 if [[ ! -d "${JCPATH}" ]]; then
-  echo "JCAGENT IS NOT INSTALLED ON THIS MACHINE."
+  echo "Jcagent is not installed on this machine."
   exit 1
 fi
 
 # Is zip installed?
-if ! which zip 2> /dev/null; then
+if ! which zip &> /dev/null; then
```
istevenson-JC: Why are we redirecting errors to `/dev/null`? I understand not wanting to have standard output display, but do we not want to display errors for troubleshooting purposes?

scripts/macos/agent_diagnostics.sh
```diff
@@ -25,47 +26,62 @@ if [[ ! -d "${JCPATH}" ]]; then
...
+    for i in "${LOGFILES[@]}"; do
+      if [[ -f "${JCLOG}""${i}" ]] || [[ -d "${JCLOG}""${i}" ]]; then
+        tar -rf "${TARFILE}" "${JCLOG}""${i}" 1> /dev/null
+        LOGIT+=("${JCLOG}${i} has been added to ${TARFILE}.")
+      fi
+    done
+  else
+    for i in "${LOGFILES[@]}"; do
+      if [[ -f "${JCLOG}""${i}" ]] || [[ -d "${JCLOG}""${i}" ]]; then
+        zip -r "${ZIPFILE}" "${JCLOG}""${i}" > /dev/null 1
```
istevenson-JC: The redirection here should be consistent with the redirection on the line above, `1> /dev/null`.

scripts/windows/agent_diagnostics.ps1
```diff
@@ -32,17 +32,19 @@ function zipjc {
...
 function ziplog {
     # Zip the log files.
-    $LOGFILE="jcagent.log"
-    cp $JCLOG\$LOGFILE $WRITEPATH\$LOGFILE
-    Compress-Archive -Path $WRITEPATH\$LOGFILE -Update -DestinationPath $WRITEPATH\$ZIPFILE
-    Remove-Item -Path $WRITEPATH\$LOGFILE
-    $ziplog_out+=("`tjcagent.log has been added to $ZIPFILE.zip")
+    $fileArray=(Get-ChildItem -Include jc*, Jump* -Name)
+    foreach($LOGFILE in $fileArray) {
+        Copy-Item $JCLOG\$LOGFILE $WRITEPATH\$LOGFILE
```
istevenson-JC: Why do we need to copy the file before adding it to the archive?

mhoneyman-jc: When attempting to write log files housed in 'c:\windows\temp' to zip with compress-archive, it fails with an error stating:

'The process cannot access the file 'C:\Windows\Temp\jcagent.log' because it is being used by another process.'

The workaround for this is to copy to the needed files to a temp location and compress from there.

scripts/macos/agent_diagnostics.sh
mhoneyman-jc: Yes it should, and now it is.

scripts/linux/agent_diagnostics.sh
mhoneyman-jc: that should have been 1>. tired eyes missed it.

scripts/linux/agent_diagnostics.sh
mhoneyman-jc: ps ax | grep jcagent
produces a result that includes the grep jcagent command.

ps ax | grep [j]cagent
produces a result without the grep jcagent command

the end result is not having to deal with the extra line of irrelevant output.

istevenson-JC: Oh, that makes sense. I didn't consider the fact that the `grep` was being applied to `ps` output.

mhoneyman-jc:
found a little issue with the output file name that I fixed since Ian's approval yesteerday.

PR #180, opened 2019-12-02
Sa 669 aws serverless

kmoorehead-jc:
First steps towards getting this published to the Serverless Application Repository is getting the source code on a publicly-available repository.

Notes: Links go to support-admin-tools repo right now. That should be identical to this repo at the moment. Will get the links up to date when this goes live. Semantic version is listed as 0.0.6. Will update to 0.1.0 when the app goes live in the Serverless Application Repository.

I also ask that you do not attempt to deploy this in us-east-1, as we have an Early Access person using that region's version. Any other region should be fine. If you would like to use a region with it already deployed, let me know and I will clean it out quickly so you can test.


kmoorehead-jc:
Got the approval from Scott to merge with master.

PR #186, opened 2019-12-09
Remove dependency on `wget`

marques-work:
The script does not need to depend on *both* `curl` and `wget`. Not only is this redundant, but this also allows the script to be run from an EdgeRouter (i.e., vanilla EdgeOS) without installing any additional packages. Not going to lie, this is the primary motivation for this PR :).

Also, `wget` is arguably less common in OOTB/vanilla installs of Linux distros as compared to `curl`, which is _nearly_ guaranteed (except for the tiniest distros).

Tangentially, this also removes some trailing whitespace.


marques-work:
@jc-rob as you appear to be the original author of this script, would you mean reviewing this one-liner change (not counting whitespace) and merging if acceptable?

marques-work:
@btstevens @jworkmanjc @scottd3v Sorry to tag you all - but you three seem to be active committers. Any chance that you can review this tiny, but sensible (IMO) PR and merge?

PR #192, opened 2020-01-03
initial commit verify script

jworkmanjc:
Added a python script to test the variables of the prestage user enrollment and postinstall script for validity. The tests are rudimentary but cover the basic set of cases where a user might enter the variables incorrectly. Additional test discovery could be useful to expand the number of tests we run before validating the scripts are valid before packaging.


jworkmanjc:
Made one change to the postinstall test to report the problematic variable in the test output and added notes to the readme file. The test suite can be run from any location as long as the users change the variables to match the location on their system. That's noted in the readme now.

PR #204, opened 2020-02-10
update to use silent flag, require connect key, do username checking

wsteidleyjc:
In order to keep people from having a bad time with their install I'd like to require the connect key to be present and check if the user is a Secure Token admin unless they use the `-s` parameter for silent (connect key still required)

### DIFF
--- a/scripts/macos/install_agent_and_serviceaccount.sh
+++ b/scripts/macos/install_agent_and_serviceaccount.sh
@@ -91,10 +91,31 @@ else
     return 1
   }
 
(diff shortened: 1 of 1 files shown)


jworkmanjc:
Looks good to me, just tested the install scenarios. I added a check for the connect key length since this branch checks for the secure token status on the current user.

scottd3v:
@wsteidleyjc are we choosing to not include logic that validates the password. The logic only validates that the same password is entered. It does not validate that this is actually the users macOS password.

PR #205, opened 2020-02-18
option to add email to jcimport template

scottd3v:


kmoorehead-jc:
Tested locally. Prompts for email. If y is selected, includes email field in sheet. That sheet can then be used to update the users' email address upon Update-UsersFromCSV.

All looks good. Approving.

PR #217, opened 2020-04-01
rename script

jworkmanjc:
Rename script to set a mac system's Hostname, LocalHostName and ComputerName to the value set in the JumpCloud Console for a given System's Display Name


scottd3v:
Minor formatting update. Approved

PR #260, opened 2020-09-03
Sa 1371 docs update

jworkmanjc:
Docs update to remove old parameter values from Get-JCSystemInsights


epanipinto-jc (approved):
good fixes

PR #263, opened 2020-09-11
Sa 1411 jcagent pppc

jworkmanjc:
PPPC profile example for jumpcloud-agent and bash where additional PPPC permissions may be necessary.


kmoorehead-jc (approved):
Got a chance to take a look at this and ask a couple questions about some pieces I wasn't clear on but it looks solid.

Approving.

PR #266, opened 2020-09-30
Adding trace

scottd3v:
Duplicate of the debug command but the variable has been updated and the name of the launch daemon


jworkmanjc (approved):
Same as Debug but just set to Trace. looks good.

PR #268, opened 2020-10-12
reinstall MDM switch

jworkmanjc:


jworkmanjc:
This update just adds an optional switch for admins to reinstall JumpCloud MDM using the same install MDM command. If the reinstall boolean is set to true, the command will reinstall the JumpCloud MDM over the existing profile.

PR #272, opened 2020-12-03
exit 1 on failed agent install

jworkmanjc:
Add an exit condition to account for scenarios where the installer throws an error.

If installer -pkg /tmp/jumpcloud-agent.pkg -target / returns 1, exit the script with exit code 1, else continue and exit 0


kmoorehead-jc (approved):
Seems pretty straightforward. At least lets the admin know that something went wrong - definitely better than not doing that.

Approving.

PR #275, opened 2020-12-22
API changed - now requires "name" + "sharedSecret"

ippikin:
The script was no longer working, the API now requiring "name" and "sharedSecret" keys for each update


dsilva-jc:
This is being handled in a more recent PR so closing this PR. The change should take effect this week.

PR #278, opened 2021-01-06
Add missing line continuation

tomi-bigpi:
The `curl` command fails due to a missing line continuation backslash.


jworkmanjc:
Hi @tomi-bigpi, we recently added a number of syntax changes to the system context API scrips including this change and examples for windows hosts. Thanks for the PR I'm going to mark this closed.

PR #283, opened 2021-01-13
SJ-2472 Update Logging for Password Expiration Prompt Scripts

dsilva-jc:
## Reason
This adds new logging to the 3 Password Expiration Prompt Scripts to catch warnings and better report expected outcomes from the following scenarios to the JumpCloud Admin Console Command Results log output:

1. No active users found on the system
2. No active JumpCloud users found on the system
3. Invalid `state` for any found JumpCloud user other than `Active` (this assists non English OS versions)
4. Any Active JumpCloud users found but without their passwords expiring.

And for the 2 scripts that validate user password expiry as part of the logic to prompt for password reset, we're also outputting:
1. When the Active JumpCloud user has their password set to never expire.

### DIFF
--- a/PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Prompt Expiring Users To Update Password.md
+++ b/PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Prompt Expiring Users To Update Password.md
@@ -101,7 +101,7 @@ bool bWait);
             'ComputerName'    = $ComputerName;
             'SessionId'       = $SessionId;
             'ResponseId'      = $Response;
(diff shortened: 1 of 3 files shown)


kruggierojc (approved):
Tested each script with the following setups: 
1. No active Users
2. Logged in as Local non-JC User
3. Logged in as Local JC User(Non-English System)
4. Logged in as Local JCUser with expiring password
5. Logged in as Local JCUser with password set to Never Expire
6. Logged in as Local JCUser(Password expiration enabled but not expiring)

Each iteration provided expected output/log results and created notifications as expected. Email notifications were also sent and received as expected as well.

PR #286, opened 2021-02-04
Sa 1613 system context windows example

jworkmanjc:
Updated syntax for mac system context example scripts, added system context example scripts for windows hosts.


kmoorehead-jc (approved):
Looks solid to me! All functions tested well.

Approving.

PR #289, opened 2021-02-22
added static fields to PUT

dsilva-jc:
Added 2 extra variables of "name" and "sharedSecret" to script to account for changes to v1 API made in Nov 2020


kyle-randolph (approved):
Looks good and works as expected.

mhoneyman-jc (approved):
works as expected.

Issue #184, opened 2019-12-07
./RADIUS_IP_Update.sh: 73: ./RADIUS_IP_Update.sh: Syntax error: "(" unexpected (expecting "fi")


Dreamurrrr:
./RADIUS_IP_Update.sh: 73: ./RADIUS_IP_Update.sh: Syntax error: "(" unexpected (expecting "fi")

wilsonsyl:
Yes, I also receive this error message.

SysAdminSeth:
I also experienced this issue.

I fixed the bug and it is currently waiting on a pull request: https://github.com/TheJumpCloud/support/pull/290

PR #290, opened 2021-03-01
Update dhcpRadiusUpdate.sh

SysAdminSeth:
Unneeded parenthesis cause script execution failure while running on Unifi Dream Machine Pro.

./jumpcloudupdate.sh: line 81: syntax error: unexpected "(" (expecting "fi")

Removing them allows for code to function as intended.

This should resolve: https://github.com/TheJumpCloud/support/issues/184


randydwhitephotography (approved):

jworkmanjc (approved):
I'm fine with this change, it's odd it's not working on some systems but I suspect that's an issue with the bash versions? 

Either way I've tested both scripts in main branch and this proposed change and they work.

gweinjc (approved):
Tested and change worked correctly

PR #293, opened 2021-03-17
Update CheckProgramInstalled to check WOW6432Node

PeetMcK:
CheckProgramInstalled is checking for the 32bit @HKLM:\Software\Microsoft\Windows\CurrentVersion\Uninstall\*. The registry keys for 32 bit uninstaller live in HKLM:SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\* 

Without the addition, the script always installs 'prereq Visual C++ 2013 x86'


PeetMcK:
¯\\\_(ツ)\_/¯

PR #295, opened 2021-03-17
Change path separator from / to \

PeetMcK:
AgentIsOnFileSystem seems to function with '/' or '\', but since your describing a windows filesystem path '\' is the standard.


PeetMcK:
¯\\\_(ツ)\_/¯

PR #299, opened 2021-04-23
SA-1863 - Remove_windows_update_scripts

btstevens:
Removes CMD based windows update scripts replaced by Windows update policy/configuration


kmoorehead-jc (approved):
Removal of all 4 Windows Update command from the command gallery.

Approving. Looks good.

PR #301, opened 2021-05-03
Sa 1824 jc deployment with sdk functions

jworkmanjc:
## Issues
* [SA-1824](https://jumpcloud.atlassian.net/browse/SA-1824) - Invoke-JCDeployment Command Failure

## What does this solve?
Invoke-JCDeployment used to reset a command type to linux when run. This fix replaces Set-JCCommand to the SDK equivalent which sets the command type correctly.

The fix isn't particularly tricky but understanding how the Invoke-JCDeployment command works in conjunction with commands is somewhat challenging. The documentation on the wiki and our api docs should be updated to note that Windows commands should contain environment variables ```${Env:$variableToReplace}``` and Mac/ Linux should contain 'normal' variables ```$variableToReplace```.


kmoorehead-jc (approved):
Documentation updates auto-generated. Update Invoke-JCDeployment to utilize Set-JcSdkCommand function.

Looks solid. Approving.

PR #302, opened 2021-05-12
Sj 2999 rename mac and windows desktop script

jworkmanjc:
## Issues
* [SJ-2999](https://jumpcloud.atlassian.net/browse/SJ-2999) - Mac - username rename script needs review, potential bug
* [SA-1892](https://jumpcloud.atlassian.net/browse/SJ-1892) - Add windows script to change background lock screen

## What does this solve?
The rename mac user script contained an invalid regex patten which would incorrectly identify users with "com" as invalid accounts. The regex pattern was updated to only filter out "com." user accounts such as launch daemons or agent associated accounts.

Added a script example for admins to change a windows 10 lock screen and/or desktop background image.

N/A


btstevens (approved):

ahendrickson-jc (approved):
Groovy.  Thanks, Joe.

PR #303, opened 2021-05-14
Update Verification script to work with MacOSMajorVersion 11

rverrips:
The original script assumes the major version for MacOS will be 10 - for the longest time this would be correct, but latest version of MacOS bumped that up to 11, and it's expected all yearly releases going forward will be major version number increments, I.e expect MacOS 12.0 later in 2021.


jworkmanjc:
Hey @rverrips, thanks for the eyes on this [we did update](https://github.com/TheJumpCloud/support/commit/5b75a76e59cbc3f508235a6db24ad5c3d82dd3c8) some of this older syntax to make things big sur (and forward compatible). I'm going to close out this PR, thank you.

rverrips:
Sure, not prob's - Happy to help where I can 👍🏼

PR #309, opened 2021-06-25
HE-1679: support script changes for log file archive changes

kwhitejc:
## Issues
* [HE-1679](https://jumpcloud.atlassian.net/browse/HE-1679) - Update log rotation size, increase rotated file count

## What does this solve?
This limits log files to 5MB, which is an attachment limit for some support systems. It also increases the number of rotated log archives we keep, to a total of 10 (primary, plus 9), for a total of 50MB of log data.

No

## How should this be tested?
For the agent, jcagent.log is still the primary log file, but rotated files will now be named jcagent.log.1, jcagent.log.2, ..., jcagent.log.9, in increasing chronological order.

### DIFF
--- a/scripts/linux/agent_diagnostics.sh
+++ b/scripts/linux/agent_diagnostics.sh
@@ -26,10 +26,10 @@ if [[ ! -d "${JCPATH}" ]]; then
 fi
 
 # Is zip installed?
-if ! which zip 1> /dev/null; then
+if ! command -v zip 1> /dev/null; then
   ZPATH="false"
 else
-  ZPATH=$(which zip)
+  ZPATH=$(command -v zip)
 fi
 
 function indent() {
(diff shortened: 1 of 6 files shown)


scripts/linux/agent_diagnostics.sh
```diff
@@ -67,15 +67,18 @@ function zipjc() {
 
 function ziplog() {
   # Zip the log files.
-  LOGFILES=("jcagent.log" "jcUpdate.log")
+  LOGFILES=(
+  "jcagent.log" "jcUpdate.log"
+  "jcagent.log.1" "jcagent.log.2" "jcagent.log.3" "jcagent.log.4" "jcagent.log.5" "jcagent.log.6" "jcagent.log.7" "jcagent.log.8" "jcagent.log.9"
```
sthurstonjc, line 72: You could use wildcards as well: `jcagent.log.*`

scripts/linux/agent_diagnostics.sh
```diff
@@ -123,9 +126,9 @@ function info_out() {
     OS=$( grep release /etc/redhat-release )
   fi
   SERVICE="jcagent"
-  STATUS=$( service ${SERVICE} status 2> /dev/null )
+  STATUS=$( service "${SERVICE}" status 2> /dev/null )
   if [[ -z "${STATUS}" ]]; then
-    STATUS=$( echo "PID  PATH" ; ps ax | grep [j]cagent | awk '{print $1" "$5}')
+    STATUS=$( echo "PID  PATH" ; pgrep -f -a "jcagent" | awk '{print $1" "$5}')
```
sthurstonjc, line 131: Didn't know this would silence SC2009 *and* work that same. Also, we need to grep for "jumpcloud-agent" now. The process name is no longer `jcagent`, and I don't know when it was.

kwhitejc: Good catch, thanks! I had this fixed on my VM, but forgot to copy it back before I pushed. Fixed.

sthurstonjc (approved):

tcavelljc (approved):

PR #314, opened 2021-08-17
HE-1791 More gracefully handle skipping creating the service account

tcavelljc:
## Issues
* [HE-1791](https://jumpcloud.atlassian.net/browse/HE-1791) - Fail agent install with no ST

## What does this solve?
The agent installer now supports a special value (=skip) in the JumpCloud-SecureToken-Creds.txt file and service account creation will be skipped rather than tried and failed.

No.

## How should this be tested?
- Running the `install_agent_and_serviceaccount.sh` script with the `-s` option.
- Using the powershell module `Mac - Install Agent and Service Account` with the `-s` option.

### DIFF
--- a/PowerShell/JumpCloud Commands Gallery/Mac Commands/Mac - Install Agent and Service Account.md
+++ b/PowerShell/JumpCloud Commands Gallery/Mac Commands/Mac - Install Agent and Service Account.md
@@ -1,6 +1,6 @@
 #### Name
 
-Mac - Install Agent and Service Account | v1.1 JCCG
+Mac - Install Agent and Service Account | v1.2 JCCG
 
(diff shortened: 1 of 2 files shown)


sthurstonjc (approved):

ldonchevjc (approved):

jworkmanjc (approved):
Nice, we needed this to support installs where the customer just needs the agent installed first before anything else. Just finished testing the install script with and without the silent install switch.

tcavelljc:
> Nice, we needed this to support installs where the customer just needs the agent installed first before anything else. Just finished testing the install script with and without the silent install switch.

Awesome, thanks for taking a look.

PR #315, opened 2021-09-08
Sa 2054   update get jc system insights help docs

gweinjc:
## Issues
* [SA-2054](https://jumpcloud.atlassian.net/browse/SA-2054) - Get-JCSystemInsights Command help should display 'current' table names

## What does this solve?
Adds a sample tab output when using the help function for Get-JCSystemInsights

No

## How should this be tested?
Verify help Get-JCSystemInsights displays correct sample output


PowerShell/Deploy/Setup-Dependencies.ps1
```diff
@@ -18,6 +18,9 @@ ForEach ($DependentModule In $DependentModules)
         if ($DependentModule -eq 'PowerShellGet'){
             Install-Module -Name $DependentModule -Repository:('PSGallery') -AllowPrerelease -RequiredVersion '3.0.0-beta10' -Force
         }
+        elseif ($DependentModule -eq 'PSScriptAnalyzer') {
+            Install-Module -Name $DependentModule -Repository:('PSGallery') -RequiredVersion '1.19.1' -Force
```
jworkmanjc, line 22: New version of PSScriptAnalyzer does not play nice with our CI pipeline, need to investigate further

.circleci/workflows.yml
```diff
@@ -87,10 +87,13 @@ workflows:
             - Initial Setup
             - Build Help and Test Files
             - Build Nuspec
+            - Invoke Git Commit
           context: aws-credentials
       - Invoke Git Commit:
           requires:
-            - Module Validation Tests
+            - Initial Setup
```
jworkmanjc, line 94: We flipped the git commit to occur before Module Validation Tests, this will write back the changelog + module banner + docs to the repo before running tests (which would always fail if a new version of PSModule was specified). Switching the order should now always write back a new changelog, module banner and docs, the Module Validation will fail on the first round but at least the dev will have the files they need to change to pass module validation tests next commit.

jworkmanjc (approved):
Looks good to go!

.circleci/workflows.yml
```diff
@@ -3,7 +3,7 @@ parameters:
   ManualBuild:
     description: 'When `true`, pipeline will continue to build even if its been filtered out by config.yml (Will Not Run on Master Branch)'
     type: boolean
-    default: false
+    default: true
```
kmoorehead-jc, line 6: @gweinjc @jworkmanjc Should this still be set to true?

kmoorehead-jc (commented):
Version and date updates. Adding info about tab-completing and the current tables available in the module.

One question waiting on being answered but other than that looks good. Will approve once we close that conversation.

.circleci/workflows.yml
kmoorehead-jc: Sounds like this shouldn't matter. Won't run on `master` branch anyway so we should be good to go.

kmoorehead-jc (approved):
Comment resolved. Approving.

PR #321, opened 2021-11-30
HE-1935 Add messaging and user confirmation for macOS full disk access

tcavelljc:
## Issues
* [HE-1935](https://jumpcloud.atlassian.net/browse/HE-1935) - Update Installer Script for macOS

## What does this solve?
When installing the agent on Monterey, the user is warned about needed to give the agent FDA and asks to confirm this unless the `-f` flag is passed in.

No.

## How should this be tested?
Run the install script with and without the `-f` option and on Monterey and an older macOS version.
* Monterey
  * with `-f`: a message about FDA is printed but no user confirmation is asked for
  * without `-f`: a message about FDA is printed and user must confirm `y` or `n`
* Any older macOS version: no FDA message is printed

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -3,7 +3,7 @@ parameters:
   ManualBuild:
     description: 'When `true`, pipeline will continue to build even if its been filtered out by config.yml (Will Not Run on Master Branch)'
     type: boolean
-    default: true
+    default: false
   EnableDebugging:
     description: 'When `true`, debugging commands will run and output to terminal.'
     type: boolean
--- a/scripts/macos/install_agent_and_serviceaccount.sh
+++ b/scripts/macos/install_agent_and_serviceaccount.sh
@@ -14,9 +14,12 @@ SECURETOKEN_ADMIN_PASSWORD=''
 # Flag for installing without checking SECURETOKEN_ADMIN_USERNAME or SECURETOKEN_ADMIN_PASSWORD
 SILENT_INSTALL=0
 
+# Flag for installing without interactive prompts
+UNATTENDED_INSTALL=0
+
 # You can also specify one or more values with parameters
 # -k YOUR_CONNECT_KEY -u SECURETOKEN_ADMIN_USERNAME -p SECURETOKEN_ADMIN_PASSWORD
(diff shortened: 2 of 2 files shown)


.circleci/workflows.yml
```diff
@@ -3,7 +3,7 @@ parameters:
   ManualBuild:
     description: 'When `true`, pipeline will continue to build even if its been filtered out by config.yml (Will Not Run on Master Branch)'
     type: boolean
-    default: true
+    default: false
```
tcavelljc, line 6: This build was failing because of an issue with the PowerShell build but that isn't needed for this. @jworkmanjc recommended disabling this build for now. There will still be a failing build on the PR but it can be ignored.

gshonlejc (approved):

scripts/macos/install_agent_and_serviceaccount.sh
```diff
@@ -173,6 +189,17 @@ else
     exit 1
   fi
 
+  # FDA is only needed on Monterey and later
+  if [[ "$MacOSMajorVersion" -ge 12 ]]; then
+    FDA_MESSAGE='JumpCloud requires Full Disk Access for our agent to properly adjust the PAM Modules for authentication. During installation, System Preferences will open, and Full Disk Access will need to be manually allowed within 5 minutes to successfully complete the installation.'
```
sthurstonjc, line 194: Suggest "configure" instead of "adjust"

sthurstonjc:
I'm wondering how we reconcile the FDA prompts with the `SILENT_INSTALL` option, but I don't have a good suggestion.

tcavelljc:
> I'm wondering how we reconcile the FDA prompts with the `SILENT_INSTALL` option, but I don't have a good suggestion.

Yep, I definitely considered that and thought about just reusing the SILENT_INSTALL option. I could honestly go either way but landed on keeping that separate.

sthurstonjc:
> 

I guess if "silent" means "no user prompts" we *could* re-use it. Is there an install scenario where we'd want to prompt the user to enable FDA but we wouldn't want to prompt them for credentials for service account creation?

sthurstonjc (approved):

PR #323, opened 2021-12-02
Sa 2268 managed apple id assignment

jworkmanjc:
## Issues
* [<SA-2268>](https://jumpcloud.atlassian.net/browse/SA-2268) - Custom script to add ManagedAppleID 

## What does this solve?
There is no UI element to add the ManagedAppleID field to a user in JumpCloud. This script should help admins, generate a template to set this field.

In the next PowerShell SDK release we'll include the field to set and get the current ManagedAppleID for users in an org. This script will have to be updated at that point in time. 

## How should this be tested?
Pull the script, generate the file, update a few users with a Managed Apple ID field

## Screenshots
Screen Shot 2021-12-02 at 10 40 22 AM


nbrownjc (approved):
This looks great! Thank you!

gweinjc (approved):
Discussed via call, everything looks good

PR #325, opened 2021-12-03
Jump cloud module 1.18.9

jworkmanjc:
## Issues
* [SA-2258](https://jumpcloud.atlassian.net/browse/SA-2258) - Fix Set-JCUser enrollment window
* [SA-2271](https://jumpcloud.atlassian.net/browse/SA-2271) - BugFix - mismatched regex in Import-JCCommand

## What does this solve?

SA-2258 - Fix Set-JCUser function where it would error when setting a user's enrollment windows anywhere between 4-7 days.

SA-2271 - Fix the Regex Pattern used in Import-JCCommand to address change in GitHub.

Add CodeArtifact support to Update-JCModule Enable the PowerShell Module to be tested from latest development versions of the PowerShell SDKs

N/A

## How should this be tested?

Automated tests account for changes. 
To test Set-JCUser, change the enrollment window for a given user it should accept an enrollmentdays parameter of 1-365
To test Import-JCCommand, verify that you can import a command within a tenant, this recently failed to work when a change was made to github.


gweinjc (approved):
Discussed via call, looks good

PR #329, opened 2021-12-17
version bump and nuspec changes

jworkmanjc:
## Issues
* [SA-2296](https://jumpcloud.atlassian.net/browse/SA-2296) - Require SDKs on new Module Install

## What does this solve?
The 1.18.11 nuspec file which controls dependencies for the powershell module on install, was missing the correct identifier on release, new installs of the JumpCloud PowerShell module would not install the required JumpCloud V1, V2 and DirectoryInsights SDKs. The 1.18.12 release should now require these additional modules on install.  

## How should this be tested?
Install the JumpCloud module on a system without the required SDKs, the required SDKs should be downloaded and installed during JumpCloud Module installation.


gweinjc (approved):

jworkmanjc:
Tested with latest version in CodeArtifact - required SDKs are now installed at PowerShell Module install time ✅

PR #332, opened 2022-01-13
Fix wrong exit code appearance caused by file downloading function

Lonedone:
## What does this solve?
* Fixes the problem that current script exits with exit code 5 although it works well, so using it this way
`powershell -command ./InstallWindowsAgent.ps1 -JumpCloudConnectKey "{{ jumpcloud_key }}"`
in, e.g., ansible, will result in error
* Speeds up a bit, since it will not have 3 seconds waiting period

I wonder why synchronous file downloading function was implemented using asynchronous function and some workarounds to make it sync

## How should this be tested?
Just try to install agent with script and check exit code after this in cmd
```
powershell -command ./InstallWindowsAgent.ps1 -JumpCloudConnectKey "{{ instant_jumpcloud_key }}"
echo Exit Code is %errorlevel%
```
At least it is reproducible on windows server 2022

## Screenshots
Screenshot 2022-01-13 at 19 07 51

### DIFF
--- a/scripts/windows/InstallWindowsAgent.ps1
+++ b/scripts/windows/InstallWindowsAgent.ps1
@@ -55,20 +55,7 @@ Function DownloadLink($Link, $Path)
 {
     [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
     $WebClient = New-Object -TypeName:('System.Net.WebClient')
-    $Global:IsDownloaded = $false
-    $SplatArgs = @{ InputObject = $WebClient
-        EventName               = 'DownloadFileCompleted'
-        Action                  = { $Global:IsDownloaded = $true; }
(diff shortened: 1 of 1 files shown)


jworkmanjc:
What's interesting is that when running the old script in powershell vs. the new suggestions you've made, both exit codes return "0".

Screen Shot 2022-01-13 at 11 45 58 AM

When running those commands through the CMD prompt, then invoking powershell I'm able to reproduce.
Screen Shot 2022-01-13 at 11 44 42 AM

Regardless I agree with your sentiment, the script should have been written this way to begin. There are a few other scripts with this same function and sleep bit. I've opened a [pull request](https://github.com/TheJumpCloud/support/pull/333) to address this on both files. I should get that merged in today. Thanks for looking at this and suggesting an edit!

PR #333, opened 2022-01-13
Async to sync downloads fix/ install windows agent

jworkmanjc:
## Issues
* [SA-2343](https://jumpcloud.atlassian.net/browse/SA-2343) - Switch Async calls to synchronous in fixWindowsAgent.ps1, InstallWindowsAgent.ps1

## What does this solve?

- Fixes the problem that current script exits with exit code 5 although it works well, so using it this way `powershell -command ./InstallWindowsAgent.ps1 -JumpCloudConnectKey "{{ jumpcloud_key }}"` in, e.g., ansible, will result in error
- Speeds up a bit, since it will not have 3 seconds waiting period

NA

## How should this be tested?

Test both functions work and exit 0

### DIFF
--- a/scripts/windows/FixWindowsAgent.ps1
+++ b/scripts/windows/FixWindowsAgent.ps1
@@ -150,7 +150,10 @@ Function InstallAgent()
 }
 Function DownloadAgentInstaller()
 {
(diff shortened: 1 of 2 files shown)


gweinjc (approved):
Tested and works for me

PR #337, opened 2022-01-20
Update Managed Apple ID script when SDKs are released

gweinjc:
## Issues
* [SA-2272](https://jumpcloud.atlassian.net/browse/SA-2272) - Update Managed Apple ID script when SDKs are released

## What does this solve?
Removes calling API manually and instead uses PowerShell SDKs

No

## How should this be tested?
* Have old version of SDK.V1 installed - ensure that it gets updated
* Don't have SDK.V1 installed - ensure that it gets installed
* Ensure that rows are skipped if they meet the following criteria:
  * Null/Whitespace, Invalid Email Address or ManagedAppleID already matches

### DIFF
--- a/scripts/automation/updateManagedAppleId.ps1
+++ b/scripts/automation/updateManagedAppleId.ps1
@@ -16,6 +16,7 @@ if (-not (Get-InstalledModule -Name JumpCloud)) {
 }
 if ([Version](Get-InstalledModule -Name JumpCloud.SDK.V1).Version -lt [Version]"0.0.27") {
     Install-Module JumpCloud.SDK.V1 -Force
+    Install-Module JumpCloud -Force
 }
 Write-Host "Connecting to JumpCloud..."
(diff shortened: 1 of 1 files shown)


scripts/automation/updateManagedAppleId.ps1
```diff
@@ -14,6 +14,9 @@ if (-not (Get-InstalledModule -Name JumpCloud)) {
     Write-Host "Installing JumpCloud PowerShell Module"
     Install-Module JumpCloud -Force
 }
+if ((Get-InstalledModule -Name JumpCloud.SDK.V1).Version -notmatch 0.0.27) {
```
jworkmanjc, line 17: If version is > than .27 this will still trigger this condition. Let's review this part

jworkmanjc (approved):
Tested:
0.0.26 SDK V1 installed -> now updates to 0.0.27
If 0.0.26 was loaded, 0.0.27 should import over and pull the correct attribute data

PR #346, opened 2022-02-10
MDM group example

jworkmanjc:
## Issues
* [SA-2395](https://jumpcloud.atlassian.net/browse/SA-2395) - Group mac systems by MDM approval status

## What does this solve?

Gathering MDM approval status is tricky in the UI, maintaining system groups membership is impossible without the API. This script groups systems by the `system.mdm.userapproved` attribute. User Approved MDM is required for specific MDM workflows .

Two groups will be created while running this script: 
`MacOS-MDM-Approved`
`MacOS-MDM-Pending`

Mac systems are queried while running this script. If they have their user approval attribute set (and are managed by JumpCloud MDM) those systems are added to the `MacOS-MDM-Approved` group. Systems without this attribute are placed in the `MacOS-MDM-Pending` group. Membership is updated each time the script is run. 

No but systems enrolled in MDM are needed to test this.

## How should this be tested?

In an org with MDM, User Approved systems, non User approved systems run the script and verify that systems are added/ removed from the correct groups. 

## Screenshots
Script output:
Screen Shot 2022-02-10 at 2 45 09 PM
Approved MDM System Group:
Screen Shot 2022-02-10 at 2 46 33 PM
Pending MDM Approval System Group:
Screen Shot 2022-02-10 at 2 47 13 PM

### DIFF
--- a/_posts/2022-02-10-Group-Systems-By-MDM-Status.md
+++ b/_posts/2022-02-10-Group-Systems-By-MDM-Status.md
@@ -1,0 +1,133 @@
+---
+layout: post
+title: Group Systems by MDM Enrollment Status
+description: Automatically group systems by MDM Enrollment status
+tags:
+  - powershell
+  - groups
+  - mdm
+---
+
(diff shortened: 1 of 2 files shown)


gweinjc (approved):
looks good

PR #354, opened 2022-03-01
SA-2372: Create variable that contains org settings

gweinjc:
## Issues
* [SA-2372](https://jumpcloud.atlassian.net/browse/SA-2372) - Save Organization Settings to env variable upon Connect-JCOnline

## What does this solve?
This will allow future functions to access an organization's settings on a global scope which can be used for manipulation or display purposes, which, in theory, will reduce overhead as the settings are stored in the variable upon using Connect-JCOnline. 

This will eliminate the need to make individual calls to gather an organization's settings per function and can instead reference the global variable

The original case for this card was to have the organization settings stored in an ENV variable. Upon further investigation, ENV variables can only be a string datatype, whereas the organization settings when queried are returned as a system.object or hashtable. 

A global variable can instead store these datatypes without any problems and all functions can access this variable as needed during a session. Upon exiting the session, the variable will no longer exist.

## How should this be tested?
```powershell
# Import the module
# Adjust the filepath as required for your system
Import-Module “/Users/USERNAME/Documents/GitHub/support/PowerShell/JumpCloud Module/JumpCloud.psd1” -force
# Connect to the JCModule
Connect-JCOnline -JumpCloudApiKey "APIKEY" -force
# Call the global variable to return org settings
$global:JCOrgSettings
```

## Screenshots
image


PowerShell/JumpCloud Module/Public/Authentication/Connect-JCOnline.ps1
```diff
@@ -154,6 +154,7 @@ Function Connect-JCOnline ()
                 # Each time a new org is selected get settings info
                 $global:JCSettingsUrl = $JCUrlBasePath + '/api/settings'
                 $global:JCSettings = Invoke-JCApi -Method:('GET') -Url:($JCSettingsUrl)
+                $global:JCOrgSettings = (Get-JcSdkOrganization -Id $env:JCOrgId).Settings
```
gweinjc, line 157: Saving these settings using a global variable instead of the initial env variable due to datatype issues

jworkmanjc (approved):
I can pull these settings, this will be helpful for future work which depends on user import behavior based on particular settings

kmaranionjc (approved):
Test ran as intended. Great work!

PR #367, opened 2022-04-12
Fix Wiki Docs Generation

jworkmanjc:
## Issues
* [SA-2499](https://jumpcloud.atlassian.net/browse/SA-2499) - Fix Wiki Docs Generation

## What does this solve?

Wiki Docs should be published on merge to main branch, this addresses the fix

## How should this be tested?

Change the CI task to commit on first step with the --dry-run parameter set on Invoke-GitCommit. This should prove the task would complete successfully. 

This pipeline run is run correctly and shows that the correct branch is being referenced when committing changes to Support.Wiki:
https://app.circleci.com/pipelines/github/TheJumpCloud/support/830/workflows/257ca67e-0560-48e5-9f4d-47aa30d722d7/jobs/3232


kmaranionjc (commented):
Nice work!

gweinjc (approved):

PR #370, opened 2022-04-27
Paginate continues even if null records are returned

gweinjc:
## Issues
* [SA-2427](https://jumpcloud.atlassian.net/browse/SA-2427) - Paginate continues even if null records are returned

## What does this solve?
If a system_group with a null system exists, the previous pagination would end once it came across the null system which would not return the full result set. The fix looks at the requests "x-total-count" to know how many results should be present

Import-JCCommand was having regex issues and that was resolved as well by making adjustments to the regex

The only way to properly test this is by using an org that has a null system in a system group, which cannot be easily replicated and would have to involve database manipulation to test thoroughly

## How should this be tested?
Import the module locally from the current branch
Run Get-JCSystemGroupMember or Get-JCUserGroupMember and validate that it is returning the correct number of results


jworkmanjc (approved):
I think this looks pretty good, it's nice to see some of the reusable code packaged up into a function. Nice work making this look clean and simple. 

We still need to update the PublishToPSGallery param to be true in order for the module to release when we merge this in.

PR #372, opened 2022-05-02
Add recoveryEmail parameter to New/Set-JCUser

rybai:
## What does this solve?
Adds the ability to set the `recoveryEmail` property added to the user object in February 2022.

The API expects `recoveryEmail` to be an object with a single property of `address`. For ease of use in the PowerShell module it is accepted as a string. In the `Set-JCUser` function I updated its value in the `$PSBoundParameters` object. I can't find any good info on downsides, if any, of updating this built in object? The other alternative was to handle it in the various conditionals but was trying to avoid repeat code if possible.

## How should this be tested?
I added the tests to the Pester files, but passing a string to the `-recoveryEmail` should work fine. I didn't do any e-mail address validation but didn't notice it elsewhere in the module so I assume we're letting the API handle that.


jworkmanjc:
Hey @rybai, 

I saw this, it mostly looks good but we'd need to do some work on our end + implement the search with Get-JCUser to bring this into the next minor version release. Let me speak with the team about it tomorrow but we should be able to help get this out there. 

Best,
Joe

rybai:
Hey @jworkmanjc!

Knew I was forgetting something - Get-JCUser! I’ll take a stab at that tomorrow if y’all don’t get to it first. Thanks for taking a look, let me know if there are any other considerations you folks look for when expanding the parameter coverage. 

Thanks,
Ryan

rybai:
Morning - I went through and added `recoveryEmail` to Get-JCUser and tried to update the Pester tests. I don't have a good dev environment available so didn't run it to validate, but all my manual testing looks good, including wildcards.

I had to handle the `recoveryEmail` conversion to an object a little differently here based on how the `Process` control logic is setup. Interested to hear if anyone has other thoughts on how these types of conflicts (recasting a parameter key/value to accommodate the API) should be handled.

Fun little project, if there are other considerations to address here let me know - I'm invested!

Thanks,
Ryan

jworkmanjc:
@rybai I just wanted to post an update and note that @kmaranionjc worked your PR this past week into the latest powershell release so I think we can close this out and say thanks for submitting the PR here, we only had to change one or two things really to get your fix in.

PR #380, opened 2022-05-16
SA-2542: Update Get-JCResults to accept new params

gweinjc:
## Issues
* [SA-2542](https://jumpcloud.atlassian.net/browse/SA-2542) - Update Get-JCResults to accept new params

## What does this solve?

- Pre-requisites for parallel implementation
- Added the method, limit and body params to the function

N/A

## How should this be tested?
The only functions that use the Get-JCResults function are:
* Get-JCSystemGroupMember
* Get-JCUserGroupMember

Validate that they return the correct number of results


kmaranionjc (approved):

jworkmanjc (approved):
Looks good for now. I'm interested to see how we'll pass in that $body param eventually whether we'll convert to json in this function or pass it in as a powershell object. Either way it doesn't matter for these two functions.

Tested w/ system groups and user groups all looks good.

PR #378, opened 2022-05-13
Sa 2523 search manager by email

kmaranionjc:
## Issues
* [SA-2523](https://jumpcloud.atlassian.net/browse/sa-2523) - SA-2523

## What does this solve?
Adds search by email to manager field for Set, Get, New-JCUser
## Is there anything particularly tricky?
N/a
## How should this be tested?
1. Create new user with -manager:` New-JCUser -username testmanageremail -firstname test -lastname manageremail -email emailtest@email.com -manager jumpclouduser@jumpcloudtest.com `
2. ` Get-JCUser -manager jumpclouduser@jumpcloudtest.com `
3. `Set-JCUser -username testmanageremail -manager newmanager@jumpcloudtest.com`
4.  Repeat step 2 with new manager

PR #381, opened 2022-05-17
Jump cloud module 1.22.0

kmaranionjc:
## Issues
* [SA-2014](https://jumpcloud.atlassian.net/browse/sa-2014) - SA-2014 - Backup-JCOrganization Permission Errors During Backup (Read-Only API Key)
* [SA-2542](https://jumpcloud.atlassian.net/browse/sa-2014) - SA-2542 - Update Get-JCResults to accept new params
* [SA-2523](https://jumpcloud.atlassian.net/browse/sa-2014) - SA-2523 - Search manager parameter by email on Get/Set/New-JCUser

## What does this solve?
* Adds search by email to manager field for Set, Get, New-JCUser
* Pre-requisites for parallel implementation added the method, limit and body params to the function
* When an admin with a read-only api key attempts to run backup-jcorganization they would have been met with an error message if they attempted to backup associations between commands -> systems, commands -> system_groups; software_apps -> systems, software_apps ->system_groups. 

## How should this be tested?
See closed PRs:
#377 
#380 
#378


PowerShell/JumpCloud Module/Public/Groups/UserGroups/Get-JCUserGroupMember.ps1
```diff
@@ -52,7 +52,7 @@ Function Get-JCUserGroupMember ()
 
                     $limitURL = "{0}/api/v2/usergroups/{1}/members" -f $JCUrlBasePath, $Group_ID
                     Write-Debug $limitURL
-                    $rawResults = Get-JCResults -Url $limitURL
+                    $rawResults = Get-JCResults -Url $limitURL -method "GET" -limit 100
```
jworkmanjc, line 55: I'd love to see the method pulled in dynamically from the function name or similar? Just a note for us to eventually think about. Maybe we can do something similar with limit?

gweinjc: I was thinking this too, however in looking at some of the Set-JC* functions, we aren't always using a PUT or a POST method, it varies as far as I can tell.

See `Set-JCSystem `and then `Set-JCSystemUser`

PR #382, opened 2022-05-19
Backup-JCOrg Read Only Support

jworkmanjc:
## Issues
* [SA-2592](https://jumpcloud.atlassian.net/browse/SA-2592) - Backup-JCOrganization Read Only Bug Fix

## What does this solve?

Several cases of backup using a read only API Key were not accounted for in the last 1.22.0 release of the PowerShell Module. This release addresses the edge case where a read only administrator running Backup-JCOrganization would be met with several 'Forbidden' messages after running the command. The associations endpoints are invalid for Commands -> Systems, System_Groups & Software_Apps -> Systems, System_Groups with read only api keys. The Traverse endpoints however are accessible and are substituted if a read only administrator runs this command. 

## How should this be tested?

In JumpCloud Module 1.22.0 Run:
`Backup-JCOrganization -Path ./ -Type System, SystemGroup, SoftwareApp, Command -Association -Debug` 
This should error with several messages

Import this development module and run:
```powershell
import-module "/Path/To/support/PowerShell/JumpCloud Module/JumpCloud.psd1" -Force
Connect-JCOnline {ReadOnlyApiKeyHere}
Backup-JCOrganization -Path ./ -Type System, SystemGroup, SoftwareApp, Command -Association -Debug
```
This should not error


gweinjc (approved):
Before fix:
image

After fix:
image

No longer erroring

PR #392, opened 2022-06-27
SA-2690: Private Function for Parallel Validation Check

gweinjc:
## Issues
* [SA-2690](https://jumpcloud.atlassian.net/browse/SA-2690) - Private Function for Parallel Validation Check

## What does this solve?
The private function is called in place of manual checks to determine if a function can be ran in parallel or not. This reduces repetitive code

N/A

## How should this be tested?

* Attempt to run the following functions in Parallel (the -debug flag should inform you if the parallel validation passed/failed):
  * Get-JCUser
  * Get-JCSystem
  * Get-JCSystemUser
  * Get-JCSystemGroupMember
  * Get-JCUserGroupMember
  * Get-JCCommandResult
* If the version of PowerShell is greater than 7 and the $parallel parameter is set to $true, then the function will continue in parallel.
* If the version is less than 7, the version of PowerShell does not support parallel functionality and a warning will be displayed to the customer, the function will continue in sequential.
* If the $parallel parameter is set to $false, the function will continue in sequential


PowerShell/JumpCloud Module/JumpCloud.psm1
```diff
@@ -12,6 +12,10 @@ Foreach ($Import in @($Public + $Private))
         Write-Error -Message "Failed to import function $($Import.FullName): $_"
     }
 }
+
+# Check to see if parallel processing is available for the session
+$global:JCParallel = Get-JCParallelValidation
```
jworkmanjc, line 17: Just a note for later, we'll be calculating this value with [SA-2727](https://jumpcloud.atlassian.net/browse/SA-2727)

jworkmanjc:
Get-JCUser test:
Screen Shot 2022-06-28 at 3 29 53 PM


Get-JCSystem test:

Get-JCSystemUser
Get-JCSystemGroupMember
Get-JCUserGroupMember
Get-JCCommandResult
If the version of PowerShell is greater than 7 and the $parallel parameter is set to $true, then the function will continue in parallel.
If the version is less than 7, the version of PowerShell does not support parallel functionality and a warning will be displayed to the customer, the function will continue in sequential.
If the $parallel parameter is set to $false, the function will continue in sequential

jworkmanjc (approved):
Tested all the cases here, It's soo much faster than before yay! 

Seriously nice work here, thanks for the pivots!

PR #396, opened 2022-07-11
Sa 2733 search filter mangaer field

jworkmanjc:
## Issues
* [SA-2733](https://jumpcloud.atlassian.net/browse/SA-2733) - Manager Search updates on Get-JCUser

## What does this solve?

Given the case of multiple users with similar emails or usernames

`jemartin`, `emartin`
`jemartin@example.com` , `emartin@example.com`

Attempt to use Set-JCUser on a random user to set their manager to emartin

Set-JCUser -id <userid> -manager emartin
Set-JCUser -id <userid> -manager emartin@example.com

An Error was thrown because Set-JCUser is trying to pass an array for the manager field (jemartin’s id and emartin’s id), which is not accepted. This PR addresses the issue and only returns exact (case-insensitive) matches

## How should this be tested?

`Tests added to Get-JCUser.tests.ps1`
The case above should no longer return an error


gweinjc (approved):
Tested changes, seems to be properly searching for users and only returning one result

PR #401, opened 2022-07-22
Sa 2725 commands gallery pipeline

jworkmanjc:
## Issues
* [SA-2725](https://jumpcloud.atlassian.net/browse/SA-2725) - Commands Gallery Pipeline

## What does this solve?

Commands Gallery files are now validated to a degree. A `commands.json` file is generated when new files or files are modified in the command gallery. Null values or values which are not compiled into the `commands.json` file throw an error in our CI pipeline. 

## How should this be tested?

Within this repository commit a new command, the commands.json file should be re-generated with the contents of the file. 
Pull the commits and verify the commands.json file contains the information from your command.
Delete your command from the gallery and commit the change to this branch. the `commands.json` file should be updated and remove file on the next automated commit.


gweinjc (approved):
Appears to be pretty straight forward to me

Can't ever have too much automation

.circleci/scripts/parse_commands_to_json.py
```diff
@@ -0,0 +1,75 @@
...
+    path = os.path.join(commandPath, part)
+    for file in os.listdir(path):
+        if file.endswith(".md") == False: continue
+
+        filePath = os.path.join(path, file)
+
+        cmd = {
+            'name': '',
+            'type': '',
+            'script': '',
```
rshawjc, line 31: I realize that after writing this and other discussions, that we should have this match these variables as close as we can to the destination (in this case, very similar to savedAgentCommands in mongo). 

The extra properties are good to go (link, description), but we should change `script` to `command` here.

kmaranionjc (approved):
Test worked as expected. Cool little pipeline

.circleci/scripts/parse_commands_to_json.py
jworkmanjc: Alright this change is done, related, I added one more automation to this workflow. 

For any script in the repository, the ```setLink.py``` script will take the contents of the Import-JCCommand -URL "someURL" and replace it with a direct link to the file as it will exist in the main branch post-merge.

This addresses two problems I realized I skipped over the weekend:
1) The URLs to the .md files would need to have been updated after we merged to master (this was our current flow)
2) In addition, four of the Git.IO short links we've used for many of these commands were no longer working. Git.IO links are depreciated so I figured it was best to link to files directly.


Any older links like this
```Import-JCCommand -URL 'https://git.io/JJL0x'```
should now be updated to point to their direct path:
```Import-JCCommand -URL "https://github.com/TheJumpCloud/support/blob/master/Linux%20Commands/Linux%20-%20Snap%20-%20Install%20Microsoft%20VS%20Code.md"```

PR #404, opened 2022-07-28
remove old commands from PSGallery

jworkmanjc:
## Issues
* [SA-2777](https://jumpcloud.atlassian.net/browse/SA-2777) - Remove Old and Redundant Commands from the PowerShell Gallery. 

## What does this solve?

These removed commands are part of the product or no longer necessary for most customers.


gweinjc (approved):
I don't see any commands that should be kept, approve

kmaranionjc (approved):
Don't see any old command we should keep

PR #406, opened 2022-08-02
Add command to install RAS for EA on Windows

wwaughjc:
## Issues
* [TX-360](https://jumpcloud.atlassian.net/browse/TX-360) - Add command to install RAS for EA on Windows

### DIFF
--- a/PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Install Remote Assist.md
+++ b/PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Install Remote Assist.md
@@ -13,7 +13,7 @@ $installerURL="https://cdn.awsstg.jumpcloud.com/TheJumpCloud/jumpcloud-remote-as
 $JumpCloudThumbprint="7A4844FBF481047BEDBB7A8054069C50E449D355"
 $installerTempLocation="C:\Windows\Temp\JumpCloudRemoteAssistInstaller.exe"
 
-Write-Host "Downloading JumpCloud Remote Assist installer now"
+Write-Host "Downloading JumpCloud Remote Assist installer"
 try {
     $ProgressPreference = "SilentlyContinue"
     Invoke-WebRequest -Uri $installerURL -OutFile $installerTempLocation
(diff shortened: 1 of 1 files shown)


PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Install Remote Assist.md
```diff
@@ -0,0 +1,81 @@
...
+        Write-Error "Failed to run JumpCloud Remote Assist installer."
+        exit 1
+    }
+
+    Write-Host "JumpCloud Remote Assist installer returned $($installerProcess.ExitCode)."
+
+    exit $installerProcess.ExitCode
+}
+finally {
+    Remove-Item "$installerTempLocation"
```
ttatarevicjc, line 67: I don't think we should be removing the $installerTempLocation this is causing an error with the shortcut to the binary in the desktop

PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Install Remote Assist.md
```diff
@@ -0,0 +1,81 @@
...
+    $authenticode = Get-AuthenticodeSignature "$installerTempLocation"
+    if ( $authenticode.Status -ne "Valid" )
+    {
+        Write-Error "No valid Authenticode signature found, aborting installation"
+        exit 1
+    }
+
+    if ( $authenticode.SignerCertificate.Thumbprint -ne $JumpCloudThumbprint )
+    {
+        Write-Error "No valid Authenticode signature found, aborting installation"
```
jpalmajc, line 38: This error message could be triggered because the code signing certificate was updated after the original expired. This means that the end user must update the thumbprint to the latest code signing certificate. Therefore, I would recommend to change this error message. The message should state that the thumbprint did not match and that the thumbprint might need to get updated to the latest code signing certificate thumbprint to work. We also should document how to get the thumbprint from the executable.

PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Install Remote Assist.md
wwaughjc: I'm not seeing the connection, but this is the sort of thing I struggle with on Windows: "Action A has no obvious causal relationship to result B." I never now how to respond :)

ttatarevicjc: I'm not seeing the connection either :/ I just noticed that removing the temporary file is resulting in not being able to open the binary ; see attached 

ras1
ras2

And if the script is updated with `Remove-Item "$installerTempLocation"` removed

ras_desktop1
ras_desktop2

fwiw, we do have other files that are JC installed which live in `C:\Windows\Temp`

szanjanizadehjc (approved):

PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Install Remote Assist.md
wwaughjc: If the thumbprint isn't what we expect, then isn't that potentially a supply chain poison? I don't think we want to encourage the admin to work around the invalid signature.

PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Install Remote Assist.md
wwaughjc: Looks like the problem has to do with whether or not the user double clicking on the icon is a "regular user" or "has admin privileges" (not 100% sure the correct terminology on Windows). Since the Remote Assist Agent EXE is in `C:\Program Files\Jumpcloud`, and that folder is correctly restricted to admin-read only, a non-admin cannot and should not be able to access the files in that folder. The current plan is to change the installed location to one where the permissions can be changed such that all users can read and execute the contents.

PR #420, opened 2022-09-15
add one more hash sign to fix CI

sarochjc:
## Issues
* [<jira-id>](https://jumpcloud.atlassian.net/browse/<jira-id>) - <jira-title>


PowerShell/JumpCloud Commands Gallery/commands.json
```diff
@@ -97,6 +97,13 @@
...
+    "link": "https://github.com/TheJumpCloud/support/blob/master/PowerShell/JumpCloud%20Commands%20Gallery/Mac%20Commands/Mac%20-%20Install%20Remote%20Assist.md",
+    "description": "Downloads and installs the JumpCloud Remote Assist app on a Mac system."
+  },
```
wwaughjc, line 106: I'm not sure we want these in commands.json. It might mean they are published to the commands gallery for all admins, which we wouldn't want for EA, we would just want select admins to know about these commands.

sarochjc: These are auto-generated by the CI, so I guess if we don't want this for all to see we can fake failure like what we just had :D

sarochjc: How do we use this commands gallery though?  How do we communicate it to the customers?

wwaughjc: Oh, I must have been mistaken then. I don't know what this repo does, sorry.

sarochjc: @pdasjc What do you think?

sarochjc: Per the [remote assist EA guide](https://docs.google.com/document/d/1M0IAK8gj21umsNseyysDq-LB1ubeoDb4FLvYwFJW8pg/edit), the admin is instructed to copy the commands right from the `Command` section in the support repo.  Hence, this PR effectively has no effect either way.  I'm still not sure how `PowerShell/JumpCloud Commands Gallery/commands.json` is being used but it seems safe to close this PR without merge, since we don't need the commands.json file updated for remote assist EA.

PR #425, opened 2022-09-30
remove command types if specified in command body

jworkmanjc:
## Issues
* [SA-2890](https://jumpcloud.atlassian.net/browse/SA-2890) - Remove command types if specified in command body

## What does this solve?

If a contributor specifies a command type in a preconfigured command, the command type was added to the body of the command through the previous automated process, this PR strips those command types before compiling our commands.json file.


kmaranionjc (approved):
Tested, did not add the commandType.

PR #426, opened 2022-10-04
fonts via commands example

jworkmanjc:
## Issues
* [SA-2887](https://jumpcloud.atlassian.net/browse/SA-2887) - Script Examples to install fonts via JumpCloud Commands

## What does this solve?

Windows & MacOS Scripts to install fonts on systems via commands

## How should this be tested?

Test on windows / macOS vms. Package several open source fonts, zip those fonts & upload to S3, try the script on both a windows and macOS system, the fonts should be installed after the command runs


gweinjc (approved):
Tests good!

One thing I thought of is, if the customer is using Google Fonts, they are publicly accessible and can be downloaded in a single CURL which would bypass the need to download and host them in an S3 instance

PR #431, opened 2022-10-12
Sa 2866 set jc command include command type

kmaranionjc:
## Issues
* [SA-2866](https://jumpcloud.atlassian.net/browse/SA-2866) - SA-2866-Set-JCCommand-include-commandtype

## What does this solve?
Fixes issue with commandType, timeout, and launchType get set to default when updating a command. Updating parameter(s) should not change other variables/objects that are not being updated.
N/A
## How should this be tested?
1. Update name of a command or any of the parameters 
2. Validate commandType, launchType, and timeout are not changed


jworkmanjc (approved):
Thanks for making those changes, this looks good approved!

PR #436, opened 2022-11-03
Sa 2938 set jc system missing description field

kmaranionjc:
## Issues
* [SA-2938](https://jumpcloud.atlassian.net/browse/SA-2938) - SA-2938-Set-JCSystem-Missing-Description-Field

## What does this solve?
Adds description parameter to Set-JCSystem and Get-JCSystem which enables system description edit and search
N/A
## How should this be tested?
1. Validate these set and get commands work for description
- Set-JCSystem -id "YOUR_SYSTEM_ID" -description "Test Description"
- Get-JCSystem -description "Test Description"


jworkmanjc (approved):
Looks good to me, the one thing we'll have to worry about in 2.1.0 is that we'll have removed module banner references. When we merge back & release we'll just want to keep that in mind and remove those files if they are in the branch.

gweinjc (approved):
Looks good to me, nice work.

See Joe's notes about the merge conflicts

PR #438, opened 2022-11-07
Sa 2670 jc backup error reporting

jworkmanjc:
## Issues
* [SA-2670](https://jumpcloud.atlassian.net/browse/SA-2670) - Improved Error Reporting for Get-JCBackup

## What does this solve?

Backup-JCOrganization creates background jobs for each type of backup object requested. If there was a timeout, internet outage or other API error which prevented the job from gathering data from the JumpCloud API an error was not thrown. Now the tasks requested of each job are wrapped in a try/catch loop. Status of each job should be reported at the end of a backup job.

## How should this be tested?

- Import the local module and connect to a JumpCloud organization.
- Ensure at least one user is bound to one system

Test that the function still works as expected
- Run `Backup-JCOrganization -Type Organizations, User -Path -Association ./ -debug`
- The Backup Jobs should complete without error

Test that the function throws an error
- Before running the command be prepared to disconnect the network halfway through the command
- Run `Backup-JCOrganization -Type Organizations, User -Path -Association ./ -debug`
- The Backup Jobs tasks should complete but error for at least several jobs. 
- The Jobs that failed should be displayed at the end of the function run

## Screenshots

Successful Job
Screen Shot 2022-11-07 at 2 41 09 PM

Job with failures
Screen Shot 2022-11-07 at 2 41 22 PM

### DIFF
--- a/PowerShell/JumpCloud Module/Public/Utilities/Backup-JCOrganization.ps1
+++ b/PowerShell/JumpCloud Module/Public/Utilities/Backup-JCOrganization.ps1
@@ -69,7 +69,9 @@ Function Backup-JCOrganization {
         ${PassThru}
     )
     Begin {
-        If ([System.String]::IsNullOrEmpty($env:JCApiKey) -or [System.String]::IsNullOrEmpty($env:JCOrgId)) { Connect-JCOnline }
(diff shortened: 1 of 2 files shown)


kmaranionjc (approved):
Tests worked. Looks good.

PR #439, opened 2022-11-09
Sa 1375 get jc object dynamic search some endpoints

kmaranionjc:
## Issues
* [SA-1375](https://jumpcloud.atlassian.net/browse/SA-1375) - SA-1375-Get-JCObject dynamic search endpoints

## What does this solve?
Fixes a bug where radius_server and cloud directory endpoints output incorrect values
## How should this be tested?
1. Run `get-jcRadiusServer -Name "Test"`. This search uses a searchTerm which will output any radius server that have the name ex. RadiusTest and JCTest will get outputted if you search for the name Test

- Multiple Servers "Name Search" | Should return "Matches" (non-case sensitive)
- Name Search with no matching searches should return nothing

3. Validate g_suite or office_365 `Add-JCAssociation`   ex. `Add-JCAssociation -Type g_suite -Name "test" -TargetType user  -TargetName "tester"`. This should add the user tester to the g_suite directory test

- Multiple Servers "Name Search" | Should return "Matches" (non-case sensitive)
- Name Search with no matching searches should return nothing


jworkmanjc:
I'm only wondering here https://github.com/TheJumpCloud/support/pull/439/files#diff-9c72d011186c4c20a9672e5b5800c9f035006719fe4f272bd05412806590a183R226

On line 226, if we want to make a special case for Radius Servers and O365/ Gsuite? 

Right now the search by name functionality is not "technically" case sensitive. It'll be case sensitive for other object types however. 

If we did anything it would be like:
If type -eq radius, o365, gworkspace; print error w/o message about the search being case sensitive

PowerShell/JumpCloud Module/Private/NestedFunctions/Get-JCObject.ps1
```diff
@@ -152,13 +197,24 @@ Function Get-JCObject {
...
                         If ($ReturnCount -eq $true) {
                             $Directory = $Result.results | Where-Object { $_.Type -eq $JCType.TypeName.TypeNameSingular }
                             $Result.totalCount = ($Directory | Measure-Object).Count
                             $Result.results = $Directory
                         } Else {
                             $Result = $Result | Where-Object { $_.Type -eq $JCType.TypeName.TypeNameSingular }
                         }
+                        # manual filter since the API doesn't support it
+                        foreach ($dirResult in $Result) {
+                            if ($dirResult.Name -eq $SearchByValueItem) {
```
jworkmanjc, line 210: Are we all okay with the search being -eq? we could get a bit more specific or less so. 

Right now for Radius Server: name `test` will return results for
"someTest", "test", "a name with test in there somewhere"

For gworkspace/ o365, name `test` will return results for:
objects named exactly "test", "TEST", "tEsT"

Would it make sense to change this work more like search term? my guess is no?

PR #442, opened 2022-11-16
SA-3009: Currently active devices not returning when querying lastContact

gweinjc:
## Issues
* [SA-3009](https://jumpcloud.atlassian.net/browse/SA-3009) - Currently active devices not returning when querying lastContact

## What does this solve?
When querying devices, if I search for devices `lastContact` after x date, devices that are currently active will not be returned, even though they have a `lastContact` date within the desired range

image

This occurs because the endpoint will only return results if the device's `active` field is `false`.

In order to return the expected results, we need to perform an `OR` filter on the `lastContact` and `active` fields

Example: 
`"or":[{"active":true},{"lastContact":{"$gt":"2022-11-15T00:00:00.0000000"}}]`

This will return devices that are currently active, as well as devices that are inactive but within the `lastContact` range.

This also only affects the "`after`" or `$gt` comparison. When searching devices in the past, it is expected to not return currently active devices

However, due to using an `OR` filter, we have to make two separate calls to gather the correct results depending if multiple fields are queried.

Example:
image

```Get-JCSystem -displayName "*CrowdStrike*" -filterDateProperty lastContact -dateFilter after -date 10/20/2022 -returnProperties active, lastContact, displayName```

In this example, we have to query for all devices that have checked in after 10/20/2022 OR are currently active, which is 3. I'm only looking for my device with the `displayName` including 'CrowdStrike', which is 1. The function then compares the two arrays for matching IDs and returns the match, which in this case is the `JC-CrowdStrike-Test` machine

## How should this be tested?

- Find all devices that have checked in since September 1st.
    - ```Get-JCSystem -filterDateProperty lastContact -dateFilter after -date 09/01/2022 -returnProperties active, lastContact, displayName```
- Find all Windows devices that have checked in since last week
    - `Get-JCSystem -os "Windows" -filterDateProperty lastContact -dateFilter after -date 11/09/2022 -returnProperties active, lastContact, displayName`

Perform any other queries in your org based on various fields to ensure that the correct results are returned

### CODE REVIEW

File: PowerShell/JumpCloud Module/Public/Systems/Get-JCSystem.ps1 [hunk 1]
```diff
@@ -326,21 +329,82 @@ Function Get-JCSystem () {
...
+                            return $null
+                        } else {
+                            # iterate through lists to look for matches, return matches
+                            $otherSystemList | ForEach-Object {
+                                if ($dateFilterList._id -contains $_._id) {
+                                    $resultsArrayList.Add($_)
+                                }
+                            }
+                        }
+                        $resultsArrayList = $resultsArrayList | Sort-Object -Property lastContact -Descending
```
jworkmanjc, line 379: When we build this object, we need to sort this again

jworkmanjc (approved):
Tests worked after the suggested line to sort

Issue #443, opened 2022-11-22
Regression: jumpcloud 2.1.0 powershell module fails to import function Get-JCObject.ps1


zaro0508:
**Describe the bug**
Importing the [Jumpcloud version 2.1.0](https://www.powershellgallery.com/packages/JumpCloud/2.1.0) powershell module fails when attempting to import `Get-JCObject.ps1` function.  Importing Jumpcloud [version 2.0.2](https://www.powershellgallery.com/packages/JumpCloud/2.0.2) module works without error therefore this seems like a serious regression. 

**To Reproduce**
1. Install Jumpcloud: `Install-Module -Name JumpCloud -Force`
2. Import Jumpcloud module: `Import-Module -Name JumpCloud -Force`

**Expected behavior**
No Errors

**Actual behavior**

```
PS C:\scripts> Install-Module -Name JumpCloud -Force

PS C:\scripts> Import-Module -Name JumpCloud -Force
C:\Program Files\WindowsPowerShell\Modules\JumpCloud\2.1.0\JumpCloud.psm1 : Failed to import function C:\Program Files\WindowsPowerShell\Modules\JumpCloud\2.1.0\Private\NestedFunctions\Get-JCObject.ps1: At
C:\Program Files\WindowsPowerShell\Modules\JumpCloud\2.1.0\Private\NestedFunctions\Get-JCObject.ps1:213 char:71
+ ...              $result = $result | Where-Object { $_ â€“ne $dirResult }
+                                                        ~~~~~~~~~~~~~~~~~~
Unexpected token 'â€“ne $dirResult }
                            }
                        }
                    }
                    # Validate results
                    If ($Result -and $Result.PSObject.Properties.name -notcontains 'NoContent') {
                        If ($SearchBy -and ($Result | Measure-Object).Count -gt 1) {
                            Write-Warning -Message:('Found "' + [string]($Result | Measure-Object).Count + '" "' + $JCType.TypeName.TypeNamePlural + '" with the "' + $SearchBy.Replace('By', '').ToLower() + '" of "'
+ $SearchByValueItem + '"')
                        }
                        $Results += $Result
                    } ElseIf ($SearchByValueItem) {
                        Write-Warning ('A "' + $JCType.TypeName.TypeNameSingular + '" called "' + $SearchByValueItem + '" does not exist. Note the search is case sensitive.')
                    } Else {
[...]

PR #444, opened 2022-11-22
SA-3026 - PowerShell module fails to import function Get-JCObject.ps1

gweinjc:
## Issues
* [SA-3026](https://jumpcloud.atlassian.net/browse/SA-3026) - PowerShell module fails to import function Get-JCObject.ps1

## What does this solve?
Replaces incorrect character to resolve a bug with importing the module

N/A

## How should this be tested?
Run `Import-Module JumpCloud`


jworkmanjc:
Tested on PowerShell v5. Resolves #443 
Screen Shot 2022-11-22 at 1 07 25 PM

jworkmanjc (approved):
Approved and tested on PowerShell version 5. Not sure how this Unicode character made it's way on Get-JCObject but this broke earlier versions of PowerShell. We test automatically on PS7 which interprets "–" as "-"

image

PR #446, opened 2022-11-22
SA-3026 - PowerShell module fails to import function Get-JCObject.ps1

gweinjc:
Issues
[SA-3026](https://jumpcloud.atlassian.net/browse/SA-3026) - PowerShell module fails to import function Get-JCObject.ps1

What does this solve?
Replaces incorrect character to resolve a bug with importing the module

Is there anything particularly tricky?
Redo of #444 

How should this be tested?
Run `Import-Module JumpCloud`


jworkmanjc (approved):
Approved, tested on PowerShell v5

Not sure why this char made it's way to the module but this is the issue. 

image

PR #447, opened 2022-11-29
identify pipeline position/ functions

jworkmanjc:
## Issues
* [SA-3036](https://jumpcloud.atlassian.net/browse/SA-3036) - Get-JCSystem | Set-JCSystem pipeline fix

## What does this solve?

When using a pipeline like: Get-JCSystem -description "something" | Set-JCSystem -description "SomethingElse" the command would throw because the param input of Set-JCSystem does not match the type of data stored for system SystemInsights state. 

i.e. 
Set-JCSystem requires a boolean value to be passed however Get-JCsystem returns the database value for SystemInsights data

Yes two private functions were added to help dynamically check pipeline positions in the future. This may not be the best approach because we will effectively modify each object before passing it along to the next item in the pipeline sequence. 

There's really only one case I know of where this is necessary now and that's passing items from Get-JCSystem to Set-JCSystem.

## How should this be tested?

Import the latest module

- On several systems in your org, set the descriptions to "someValue" 
- Run `Get-JCSystem -description "SomeValue"` this should return several systems
- Then run `Get-JCSystem -description "SomeValue" | Set-JCSystem -description "SomethingElse"`
- That command should throw the error that Get-JCSystem requires a boolean param

Import this Development Module 

- On several systems in your org, set the descriptions to "someValue" 
- Run `Get-JCSystem -description "SomeValue"` this should return several systems
- Then run `Get-JCSystem -description "SomeValue" | Set-JCSystem -description "SomethingElse"`
- That should not throw an error and update the description on all systems

## Screenshots

Screen Shot 2022-11-29 at 3 17 56 PM


kmaranionjc (approved):
Great job on the fix! Tests ran well.

gweinjc (approved):
Tests good. Nice use of private functions, we might be able to do some cool stuff with those in the future!

PR #467, opened 2023-02-15
SA-3184 - Radius Cert Repo Cleanup

gweinjc:
## Issues
* [SA-3184](https://jumpcloud.atlassian.net/browse/SA-3184) - Radius Cert Repo Cleanup

## What does this solve?
The purpose of this is to reduce the clutter within the file structure and to break out functions into their individual files to make future work easier to find/understand

A lot of the scripts and functions make use of the `$PSScriptRoot` variable, which references the path of that individual script run. This was fine when everything was under the same folder, however; it does not work when the functions are relegated to their own individual files.

To fix this, a global variable is created when `Start-JCDeployment.ps1` is executed that saves the path of `../support/scripts/automation/radius`

There was also an issue when generating the UserCerts that relied on the variables set in Config.ps1. This was remedied by calling `. "$JCScriptRoot/config.ps1"` in the begin block of `Generate-UserCert.ps1`

## How should this be tested?

1. Pull Branch
2. Edit `Config.ps1` and update variables for your org
3. Run `./Start-RadiusDeployment.ps1`
4. Ensure the following:
    * Root Certs are generated correctly (`../Radius/Cert`)
    * User Certs are generated correctly in their proper folder (`../Radius/UserCerts`) 
    * users.json file is generated correctly
    * Commands are generated correctly and the correct files are uploaded to them
    * Reporting is accurate for your environment

[SA-3184]: https://jumpcloud.atlassian.net/browse/SA-3184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jworkmanjc (approved):
Screen Shot 2023-02-15 at 12 42 35 PM

Thanks for cleaning this up, looks good in my test!

scripts/automation/Radius/Start-RadiusDeployment.ps1
```diff
@@ -5,22 +5,25 @@ Connect-JCOnline $JCAPIKEY -force
...
+Import-Module "$JCScriptRoot/Functions/JCRadiusCertDeployment.psm1" -DisableNameChecking -Force
 
 # Show user selection
 do {
     Show-RadiusMainMenu
     $selection = Read-Host "Please make a selection"
     switch ($selection) {
         '1' {
-            . "$psscriptroot/Generate-RootCert.ps1"
+            . "$JCScriptRoot/Functions/Public/Generate-RootCert.ps1"
```
jworkmanjc, line 20: Can't we just call `Generate-RootCert.ps1` here now that you've imported the module?

kmaranionjc (approved):
Tests looks good

scripts/automation/Radius/Start-RadiusDeployment.ps1
gweinjc: The .psm1 file only imports the individual functions, whereas these .ps1 files act more like a main rather than individual functions

PR #476, opened 2023-03-03
SA-3199 Specify macOS SSID when importing user certificate

jworkmanjc:
## Issues
* [SA-3199](https://jumpcloud.atlassian.net/browse/SA-3199) - macOS Specify SSID and associate user certificate

## What does this solve?

In the previous release, macOS end users were prompted no less than three times to set their radius certificate identity preference to the desired Radius Network, to allow eapolclient access to sign using the private key from the cert and finally allow eapolclient access to use the private key from the user cert.

radius user interaction macOS

This release eliminates unnecessary user interaction but automatically assigning the user certificate to the SSID of the radius network and giving `eapolclinet` access to the private key during certificate import.

radius user interaction macOS (1)

## How should this be tested?

Populate the `$NETWORKSSID` variable in `config.ps1`. Generate user certs and distribute to devices. When connecting to the Radius network specified in the `config.ps1` file, the System UI should only prompt for password once, and if the user selects "Always Allow" it will not prompt during the duration of the certificate's life. 

Subsequently installed certificates have a different private key and users will be prompted to enter their password when certificates are replaced.

## Screenshots

https://user-images.githubusercontent.com/54448601/222821964-e7eb6418-67ad-440f-b863-dc6b12b92f54.mov

[SA-3199]: https://jumpcloud.atlassian.net/browse/SA-3199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/automation/Radius/Config.ps1
+++ b/scripts/automation/Radius/Config.ps1
@@ -8,6 +8,8 @@ $JCUSERGROUP = 'YOURJCUSERGROUP'
 $JCUSERCERTPASS = 'secret1234!'
 # USER CERT Validity Length (days)
 $JCUSERCERTVALIDITY = 90
+# NETWORK SSID (optional)
+$NETWORKSSID = 'YOUR_SSID'
 # OpenSSLBinary by default this is (openssl)
 # NOTE: If openssl does not work, try using the full path to the openssl file
 # MacOS HomeBrew Example: '/usr/local/Cellar/openssl@3/3.0.7/bin/openssl'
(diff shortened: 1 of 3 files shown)


kmaranionjc (approved):
Looks good. That's a nice little efficiency bump

gweinjc (approved):
Looks good, should we specify that the SSID portion only applies to Mac systems?

PR #480, opened 2023-03-20
SA-3230: Replace IWR Calls with SDK Functions

gweinjc:
## Issues
* [SA-3230](https://jumpcloud.atlassian.net/browse/SA-3230) - Replace IWR Calls with SDK Functions

## What does this solve?
The PowerShell module has a 30 second timeout period with all `Invoke-WebRequest` calls. If one of these calls is an invalid request, a customer could experience a delay of 30 seconds when attempting to do basic tasks, in this instance, add or remove users/systems from a group.

If a user is already a member of a group and the Add-JCUserGroupMember function is called to add that user, a 30 second timeout will occur because that user already belongs to the group

If a user does not belong to a group and the Remove-JCUserGroupMember function is called to remove that user, a 30 second timeout will occur because that user cannot be removed because they do not exist as a member.

N/A

## How should this be tested?
1. Add a user to a JC Group
2. Attempt to add that user to the same group again - results should state that the user is already a member and the TTC should be less than 30 seconds

1. Remove a user from a JC Group
2. Attempt to remove that user from the same group again - results should state the user is not found and the TTC should be less than 30 seconds.

[SA-3230]: https://jumpcloud.atlassian.net/browse/SA-3230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jworkmanjc (approved):
Tested Add/ Remove functions for User/ System. All looks good to me!

Screen Shot 2023-03-21 at 11 01 27 AM
Screen Shot 2023-03-21 at 11 00 14 AM

kmaranionjc (approved):
Tests looks good!

PR #482, opened 2023-04-06
BAN-1163: enable disable wayland

asinghjc:
## Issues
* [BAN-1163](https://jumpcloud.atlassian.net/browse/BAN-1163) -  enable disable wayland windowing system

## What does this solve?
This script enables or disables wayland windowing system.
Remote-assist application does not work with wayland windowing system.
windowing system will need to be switched to x11.

## Screenshots

[BAN-1163]: https://jumpcloud.atlassian.net/browse/BAN-1163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/linux/wayland_enable.sh
```diff
@@ -0,0 +1,46 @@
...
+################################################################################
+
+if [[ "${UID}" != 0 ]]; then
+    (>&2 echo "Error:  $0 must be run as root")
+    exit 1
+fi
+
+enable_wayland() {
+    case ${linuxType} in
+        ubuntu|\"rocky\"|\"centos\")
```
spairajc, line 13: why we differentiating between these and debian ?

scripts/linux/wayland_enable.sh
```diff
@@ -0,0 +1,46 @@
...
+            `sed -i 's/WaylandEnable=.*/#WaylandEnable=false/' /etc/gdm3/custom.conf`
+            ;;
+        debian)
+            `sed -i 's/WaylandEnable=.*/#WaylandEnable=false/' /etc/gdm3/daemon.conf`
```
spairajc, line 17: Shouldn't this here be true ?

scripts/linux/wayland_enable.sh
asinghjc: the files in which we need to make the change are different for ubuntu|\"rocky\"|\"centos\" (/etc/gdm3/**_custom.conf_**) and debian (/etc/gdm3/**_daemon.conf_**).

scripts/linux/wayland_enable.sh
asinghjc: Not sure, the comment given in the conf file says that the line **_WaylandEnable=false_** need be commented and uncommented for the wayland enabling and disabling.

spairajc (approved):

PR #484, opened 2023-04-20
SA-3282: localUserAccount field check

gweinjc:
## Issues
* [SA-3282](https://jumpcloud.atlassian.net/browse/SA-3282) - localUserAccount field check

## What does this solve?
Previously, if a user had their localUserAccount field changed in the console, it would not be taken into account when the certs are generated which could lead to issues installing the script.

We simply added a conditional check to see if the `systemUsername` field (which corresponds with the `localUserAccount` field in the console) is not empty. If it is empty, we will continue to use the user's JumpCloud `username`. If it is not empty, we will instead use the user's `systemUsername` and set that as the user's `username` in a customObject that is returned in the `Generate-UserCert.ps1` function

No

## How should this be tested?

1. Change a user's localUserAccount is the console
2. Assign the user to a device
3. Add the user to the Radius group
4. Generate a root cert for testing
5. Generate the user certs for testing
6. Attempt to deploy the certs while the user is logged into the device

## Screenshots
image
image
image
image

[SA-3282]: https://jumpcloud.atlassian.net/browse/SA-3282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/automation/Radius/Changelog.md
+++ b/scripts/automation/Radius/Changelog.md
@@ -1,0 +1,30 @@
+## 1.0.1
+
+Release Date: April 21, 2023
+
+#### RELEASE NOTES
+
+```
+Users with Local User Account names that differ from their JumpCloud Username are now supported
+```
+
+#### FEATURES:
+
+- If a user's local account name (systemUsername) is specified on an account, the certificates for those users will be generated with their local account name (not username) and installed correctly.
+
+## 1.0.0
+
+Release Date: March 21, 2023
+
+#### RELEASE NOTES
+
+```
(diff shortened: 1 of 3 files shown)


jworkmanjc (approved):
As long as we make that change to the line as code suggested, I'm good with this release. Tested those changes locally and was able to distribute cert to a user w/ local system account.

kmaranionjc (approved):
Looks good on my side. Great job

PR #486, opened 2023-05-02
SA-3322: Radius Cert Secret Fix

gweinjc:
## Issues
* [SA-3322](https://jumpcloud.atlassian.net/browse/SA-3322) - Radius Cert Secret Fix

## What does this solve?
Fixed an issue with the `$JCUSERCERTPASS` not being correctly passed into Windows device commands when changed from default in `config.ps1`

N/A

## How should this be tested?

1. Change the default value of `$JCUSERCERTPASS` in `config.ps1`
2. Generate and Distribute certs
3. Validate in the command that the Password value is as expected
4. Ensure that certificates are installed correctly

1. Change the password value in the JC Command to a random value and rerun command
2. Should produce an error stating the password does not match

## Screenshots
N/A

[SA-3322]: https://jumpcloud.atlassian.net/browse/SA-3322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/automation/Radius/Changelog.md
+++ b/scripts/automation/Radius/Changelog.md
@@ -1,3 +1,17 @@
+## 1.0.2
+
+Release Date: May 2, 2023
+
+#### RELEASE NOTES
+
+```
+Fixed an issue with the JCUSERCERTPASS not being correctly passed into Windows devices when changed from default
+```
+#### FEATURES:
+
+- JCUSERCERTPASS was not being correctly referenced when the device commands are generated resulting in certificates not being installed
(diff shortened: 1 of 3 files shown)


jworkmanjc (approved):
Tested, with two users one with a "local Account username" one without, and updating the `$JCUSERCERTPASS` variable to something outside the default. 

Both users received their user certs and would error when the password was changed (verify that the password error now throws). 
Screen Shot 2023-05-02 at 2 14 23 PM

PR #492, opened 2023-05-22
SA-3170: Replace Get-JCCommandResults with Search-JCSdkCommandResult

gweinjc:
## Issues
* [SA-3170](https://jumpcloud.atlassian.net/browse/SA-3170) - Replace Get-JCCommandResults with Search-JCSdkCommandResult
* [SA-3345](https://jumpcloud.atlassian.net/browse/SA-3345) - PFX Permission adjustment on MacOS distribute script

## What does this solve?
This change will reduce the time it takes to aggregate command results in tenants with large amounts of existing command results. The search endpoint doesn't require as many API calls to paginate the data.

Added chmod to the MacOS distribute script due to reports of permission issues on deployment

The command result clean up had to be adjusted to account for the new return body that the SDK uses compared to the Get-JCCommandResult function.

## How should this be tested?
Attempt the deployment scripts as usual
View the command results by selection option 4 in the main menu
When viewing the command results, ensure that the latest round of command results are displayed and that the previous command results are removed in the console

[SA-3170]: https://jumpcloud.atlassian.net/browse/SA-3170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SA-3345]: https://jumpcloud.atlassian.net/browse/SA-3345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/automation/Radius/Changelog.md
+++ b/scripts/automation/Radius/Changelog.md
@@ -1,3 +1,17 @@
+## 1.0.3
+
+Release Date: May 24, 2023
+
+#### RELEASE NOTES
+
+```
+Fixed an issue affecting permissions on certain MacOS devices when attempting to deploy certs
+Improved performance when reviewing Command Results by changing fetch requests to Search-JCSDKCommandResult endpoint
+```
+#### FEATURES:
+
+- Fixed an issue affecting permissions on certain MacOS devices when attempting to deploy certs
+- Improved performance when reviewing Command Results by changing fetch requests to Search-JCSDKCommandResult endpoint
 ## 1.0.2
 
 Release Date: May 2, 2023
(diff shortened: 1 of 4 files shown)


jworkmanjc (changes requested):
I've tested these changes this morning, I think it's working correctly and I'm getting the expected results as I distribute these new commands.

No issues for me installing the certs after updating the file permissions. Overall this seems much faster given that my org has 100s of command results to sort through.

The only change I think we need to make is replacing the `Get-JCCommandResult` call from `Invoke-CommandsRetry.ps1` file to fulfill the AC of the card.

Screen Shot 2023-05-23 at 8 29 47 AM

kmaranionjc (approved):
Code and tests looks

PR #502, opened 2023-06-29
local user fix for cert CN identification

jworkmanjc:
## Issues
* [SA-3225](https://jumpcloud.atlassian.net/browse/SA-3225) - Local Username Certificate CN Fix

## What does this solve?

Previously when generating certificates with users who also have a systemUsername (local username account field set). The resulting certificates were generated with a users 'systemUsername'. Radius authentication only matches a user's username and user's with these certificates would never be allowed to access a radius backed network. 

This change tracks a user's localUsername and username (if both exist). Certs are generated with the users username but installed to the localUsername directory.

## How should this be tested?

- Associate a user who has a systemUsername (local username account field set) to a radius access group
- Generate a radius user certificate for this user
- distribute the radius certificate to the user's macOS and windows system
- the certificate should install
- the certificate should allow a user to access a radius backed network even though their localUsername is different from the CA in the resulting certificate.

## Screenshots

[SA-3225]: https://jumpcloud.atlassian.net/browse/SA-3225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/automation/Radius/Changelog.md
+++ b/scripts/automation/Radius/Changelog.md
@@ -1,3 +1,17 @@
+## 1.0.5
+
+Release Date: July 3, 2023
+
+#### RELEASE NOTES
+
+```
+Addressed an issue generating certificates for users with localUsernames (systemUsernames specified in the JumpCloud console). These user certificates were generated with the localUsername instead of their username field. The resulting certificate would never be allowed to access a radius backed network as their localUsername does not match the username.
+```
+
+#### Bug Fixes:
+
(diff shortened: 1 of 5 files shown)


jworkmanjc:
Tested both Windows/ macOS deployments. 

Both users have a local username, the installation scripts correctly install the certificate, certificate auth to radius network will not throw handshake failure.

kmaranionjc (approved):
Tests looks good

gweinjc (approved):
Certs were installed correctly for the localUser and was able to authenticate using their username

PR #507, opened 2023-08-03
Update jc agent install script

kmaranionjc:
## Issues
* [SA-3465](https://jumpcloud.atlassian.net/browse/SA-3465) - SA-3465-Update-Install-Windows-Agent-Script-Install Validation

## What does this solve?
Fixes the issue incorrect binary/exe path which was giving wrong validation success results.
N/A
## How should this be tested?
1. Run the script on a machine with no JumpCloud installed on it
2. Validate that the result output are correct
3. Validate JumpCloud is installed

### DIFF
--- a/scripts/windows/InstallWindowsAgent.ps1
+++ b/scripts/windows/InstallWindowsAgent.ps1
@@ -10,67 +10,22 @@ Param (
 #--- Modify Below This Line At Your Own Risk ------------------------------
 
 # JumpCloud Agent Installation Variables
-$msvc2013x64File = 'vc_redist.x64.exe'
-$msvc2013x86File = 'vc_redist.x86.exe'
-$msvc2013x86Link = 'https://download.microsoft.com/download/0/5/6/056dcda9-d667-4e27-8001-8a0c6971d6b1/vcredist_x86.exe'
-$msvc2013x64Link = 'https://download.microsoft.com/download/0/5/6/056dcda9-d667-4e27-8001-8a0c6971d6b1/vcredist_x64.exe'
 $TempPath = 'C:\Windows\Temp\'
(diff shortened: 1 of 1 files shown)


scripts/windows/InstallWindowsAgent.ps1
```diff
@@ -79,11 +34,14 @@ Function DownloadAndInstallAgent(
...
 
-    }
-    If (CheckProgramInstalled("Microsoft Visual C\+\+ 2013 x64") -and CheckProgramInstalled("Microsoft Visual C\+\+ 2013 x86") -and AgentIsOnFileSystem) {
-        Write-Output 'JumpCloud Agent Installer Completed'
-    } Else {
-        Write-Output 'JumpCloud Agent Installer Failed'
+        # Check if Agent is on filesystem
+        Start-Sleep -Seconds 10 # Wait 10 seconds for the agent install to complete before checking if the agent is on the filesystem
+
+        If (Test-Path -Path "$($AGENT_PATH)\$($AGENT_BINARY_NAME)") {
```
mmihevcmm, line 40: A better check would be to us sc.exe to check if the agent is running as a service.  Sometimes the agent is laid down but something went wrong during installation and the service didn't get registered.  This can be reproduced pretty easily but installing with the wrong connect key.

kmaranionjc: Added a check for the service using Get-Service, equivalent to sc.exe query

jworkmanjc (approved):
Screen Shot 2023-08-04 at 2 44 05 PM
Screen Shot 2023-08-04 at 2 20 47 PM

Looks like the updated script is calling the installer correctly now. When I pass a bogus connect key into the installer I do get the message that the agent was not installed correctly. 

A valid connect key will install the agent ✅

gweinjc (approved):
Tests good in my environment

PR #513, opened 2023-09-08
SA-3530: CrowdStrike Command Install Token

gweinjc:
## Issues
* [SA-3530](https://jumpcloud.atlassian.net/browse/SA-3530) - CrowdStrike Command Install Token

## What does this solve?
If a customer has `Require Tokens` enabled in their CrowdStrike organization to install the sensor, currently, the commands do not provision correctly and the installation fails.

With this change, the customer can provide their InstallToken to the command which will then install and provision the sensor correctly for their organization

`Require Token` needs to be enabled in the CrowdStrike organization under [Host setup and management > Deploy > Installation tokens](https://falcon.us-2.crowdstrike.com/hosts/installation-tokens)

## How should this be tested?

1. Enable `Require Token` in [Host setup and management > Deploy > Installation tokens](https://falcon.us-2.crowdstrike.com/hosts/installation-tokens)
2. Copy the Install Token
3. Pull this branch and copy the respective command into your commands (Windows or Mac)
4. Assign the Install Token to the `CSInstallToken` variable for your command
5. Run the command and monitor the results
6. The CrowdStrike sensor should be installed and the device should be appearing in the CrowdStrike host dashboard

[SA-3530]: https://jumpcloud.atlassian.net/browse/SA-3530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jworkmanjc (approved):
I can note that for both mac and windows when you pass in the correct provision token, the software is installed and the agent does show up in crowdstrike. 

If you pass an incorrect token into the installer, the software is installed, likely licensed but it'll never show up in CrowdStrike. 

I think there's a chance for customer confusion if this ever were to happen but if they were passing the incorrect value in, all they'd need to do is send a JumpCloud command to run `sudo /Applications/Falcon.app/Contents/Resources/falconctl provisioning-token NewToken` to update the value; the agent would appear in the console after.

For what it's worth both windows and mac agents installed correctly with this token addition

kmaranionjc (approved):

PR #516, opened 2023-09-25
1.0.6 multi ssid and non-exportable macos certs

jworkmanjc:
## Issues
* [SA-3558](https://jumpcloud.atlassian.net/browse/SA-3558) - Multiple SSIDs with spaces

## What does this solve?

In previous versions of the radius example scripts, multiple SSIDs could be specified as long as those SSIDs did not have a space in the name. Some SSID set like "network_5G network_2.5G" would have been valid in a previous release whereas "network 5G network 2.5G" would not. In this release, space is no longer a delimiter to split network names, instead, the semicolon ";" is used to split network names. In this release setting the `$NETWORKSSID` variable to "network 5G;network 2.5G" would assign a macOS user's wifi certificate to both networks `network 5G` and `network 2.5G`

In addition this release changes the import certificate macOS security commands slightly by adding the `-x` flag which prevents the certificates from being exported from the keychain after installation.

NA

## How should this be tested?

In the radius `config.ps1` file, change the `$NETWORKSSID` variable to a string with several network names one of which should have a space in the name. 

Generate and distribute a set of certs to a test user on a VM, when the certificate is distributed and installed on the user's device, the certificate should be set to automatically apply when connected to each network specified in the `$NETWORKSSID` variable. 

Furthermore the installed certificate should no longer be exportable once it's in the user's keychain

## Screenshots

In this screenshot both networks `TP-Link_3832` and `Some network with a space` were set to use the installed certificate for authentication after the certificate was installed: 

Screenshot 2023-09-25 at 1 43 43 PM

After the certificate was installed, it can no longer be exported by right-clicking the private key and selecting export. The following error should be displayed when a user attempts to export this installed certificate. 

Screen Shot 2023-09-12 at 10 50 38 AM 2

[SA-3558]: https://jumpcloud.atlassian.net/browse/SA-3558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/automation/Radius/Changelog.md
+++ b/scripts/automation/Radius/Changelog.md
@@ -1,4 +1,4 @@
-## 1.0.5
+## 1.0.6
 
 Release Date: September 25, 2023
 
--- a/scripts/automation/Radius/Changelog.md
+++ b/scripts/automation/Radius/Changelog.md
@@ -1,5 +1,19 @@
 ## 1.0.5
 
+Release Date: September 25, 2023
+
+#### RELEASE NOTES
+
+```
+Certificates distributed to macOS device are now imported using the -x flag to prevent them from being exported.
+```
+
+#### Bug Fixes:
+
+- For users with multiple SSIDs where one SSID has a space in the name, previous versions of the script could not account for this. This version addresses this change by passing text with a ';' delimiter rather than a space.
+
+## 1.0.5
+
 Release Date: July 20, 2023
 
 #### RELEASE NOTES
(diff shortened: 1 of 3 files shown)


scripts/automation/Radius/Changelog.md
```diff
@@ -1,5 +1,19 @@
 ## 1.0.5
```
jworkmanjc, line 1:
```suggestion
## 1.0.6
```

gweinjc (approved):
Works in my test org, nice job

kmaranionjc (approved):
Works great with my test. Awesome job!

PR #536, opened 2023-10-17
Sa 3608 build pack nuspec

jworkmanjc:
## Issues
* [SA-3608](https://jumpcloud.atlassian.net/browse/SA-3608) - nuspec/ nupkg updates for GitHub Actions

## What does this solve?

Linux runners process the `BuildNuspecForPsd1.ps1` file ever so slightly different from Windows or macOS systems. When building a nuspec file in linux, the nuspec file contains a flattened set of functions. In other words functions that are split between `public` and `private` directories were flattened into their sub directories like `users` and `systems`.

Some public functions reference these private functions by relative paths and when the nupkg file contains flattened function directories, those functions were inaccessible.

This release addresses the issue of building the nuspec file specifically for the `ubuntu-latest` runners in Github Actions. There is a new job introduced to the release step that will validate that the "Public" and "Private" directory exist within the nupkg file. The newly create nupkg file is uploaded as an artifact.

This step occurs before the manual approval job. SAs should be able to validate the nupkg file is correct before approving. 

After manual approval, the nupkg artifact is downloaded, and published to PSGallery.

### DIFF
--- a/.github/workflows/build-pack-pwsh-ci.yml
+++ b/.github/workflows/build-pack-pwsh-ci.yml
@@ -49,12 +49,12 @@ jobs:
         with:
           name: jumpcloud-module-nupkg
       - shell: pwsh
-      run : |
-        # add nuget source:
-        dotnet nuget add source "https://www.powershellgallery.com/api/v2/package" --name PSGallery
-        # get nupkg artifact:
-        $nupkgPath = (Get-ChildItem -Path:("./*.nupkg")).FullName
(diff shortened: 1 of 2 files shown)


gweinjc (dismissed):
Looks good to me

kmaranionjc (dismissed):

kmaranionjc (approved):

gweinjc (approved):

PR #544, opened 2023-11-15
Added fix to string value error

kmaranionjc:
## Issues
* [SA-3679](https://jumpcloud.atlassian.net/browse/SA-3679) - SA-3679-PolicytoPatch-Copy

## What does this solve?
Fixes issue with policy value converting a string to an integer error
n/a
## How should this be tested?
1. Run this with the current module: Set-JCPolicy -id POLICYID -AUTO_INSTALL_SCHEDULE ScheduledInstallFourthWeek
* This should output an error
2. Run the same code with this module, validate that the issue is fixed
## Screenshots

[SA-3679]: https://jumpcloud.atlassian.net/browse/SA-3679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jworkmanjc (changes requested):
I've at least validated this locally. Can you add a test to this release to configure a windows update policy + set the specific config field where we encountered this issue. Test should do this:

Create a new policy
Set the specific field where we encountered this issue in the first place
test that the setting of that field | Should -Not Throw
Validate that the field is set to the desired value 
ie. 
($policy.Values | where {$_.configFieldID -eq 1234 }).value | should -be "theExpectedValueSetAbove"

I'd want to see this as part of the release before I approve.

jworkmanjc (approved):
Tested locally and can see that this is working for the windows update policy. Thanks for the work here and looks like you have the functionality covered in your new test. ✅

gweinjc (approved):

PR #559, opened 2024-02-05
Cert Generation Tests + SN in Commands

jworkmanjc:
## Issues
* [SA-3832](https://jumpcloud.atlassian.net/browse/SA-3832) - Cert Generation SN in Commands

## What does this solve?

Changes in #553  modified the command generation process. The Serial Number variable was no longer being passed into the generated command and, certificate installations would fail on devices. This addresses the issue by passing in the serial number when generating commands.

Additional tests were written to:
- Validate a certificate serial number is correctly passed into generated macOS/ Windows commands
- When changing certificate generation types
- - The subject headers of the certificate are correct
- - Any old .crt file is deleted for a user. i.e. users should not have more than one .crt file at any given time if the certificate generation type is changed from `usernameCN` to `emailSAN` 

## How should this be tested?

No manual tests required, the functionally of the PR should be tested automatically with pester tests.


kmaranionjc (approved):
Looks good. Don't see any issue with the changes.

gweinjc (approved):
Looks good to me, nice work with the tests!

PR #578, opened 2024-05-28
CUT-4045: New staged users needing exclusionDays instead of exclusionUntil

gweinjc:
## Issues
* [CUT-4045](https://jumpcloud.atlassian.net/browse/CUT-4045) - New staged users needing exclusionDays instead of exclusionUntil

## What does this solve?
When creating a new staged user and wanting to enable MFA, instead of setting the `exclusionUntil` field for that user, the `exclusionDays` field needs to be set instead.

image

Currently in the `New-JCUser` function, we have an `enrollmentDays` parameter that allows for the setting of this field for users. This will typically work UNLESS that user's `state` is `STAGED`.

The fix for this was to include a check for the `state` field being `STAGED`, and if it is - do not set the `exclusionDays` and instead set the `enrollmentDays` field.

You do not have to explicitly set the `enrollmentDays` as it will default to 7 days. This is partly one of the reason that the issue was being presented

## How should this be tested?

Run the following command: 
```powershell
New-JCUser -firstname "brand" -lastname "newUser" -username "brand.new.user" -email "brand.new.user@alderaan.org" -enable_user_portal_multifactor $true -state "STAGED"
```

The user should be created with no errors

[CUT-4045]: https://jumpcloud.atlassian.net/browse/CUT-4045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kmaranionjc (approved):
Tested and working as expected. Nice work!
image

PR #580, opened 2024-05-29
Track installed user certificates using system insight filters

jworkmanjc:
## Issues
* [CUT-4090](https://jumpcloud.atlassian.net/browse/CUT-4090) - Track installed certificates with system insight filters

## What does this solve?

At the cost of some upfront API requests, this release tracks whether or not a certificate is considered to be installed on a given system. This change adds yet another piece of data to our `users.json` table when the global variables are first updated. For each user in the membership list we are now performing a check to identify whether or not the SHA1 value of the user's current certificate exists on the end user's device(s). If that cert is installed, the `users.json` file will be updated with a list of data under the `deploymentInfo` property. 

At the end of the day, this release adds some upfront API calls when loading the automation script but reduces the number of times we'd potentially issue a command to a device that **already** has the required certificate. 

**Changes to Users.json**

While the app is loading, the system insights certificate API is queried for Windows and macOS systems that have a given user's certificate installed. That data is tracked under the `deploymentInfo` object in `users.json`

Screenshot 2024-05-30 at 2 22 33 PM

Screenshot 2024-06-03 at 4 41 29 PM

**Viewing Deployment Info**

Instead of querying the API for command info, the monitor menu now provides access to a new function that will report on systems with the certificates installed: 

Screenshot 2024-05-30 at 2 24 35 PM

This view only requires the data cached locally when the app is run. Data is returned quickly and provides administrators with a quick at-a-glance view of the systems that have / do not have the required certificate installed. 

**Commands Distribution**

When distributing certificates, the `user.json` data is now queried to determine which users require work to be done. 

Screenshot 2024-05-31 at 12 44 16 PM

When selecting options from the distribution menu, the tool will consult the system insights record to see which certificate has been installed on a user's device. 

Selecting option 2 to distribute certs to users who do not have their certificate installed already will trigger commands on all systems that do not have the required cert installed. This applies to users who have certificates installed on other devices. New to this release, if devices already have their required user cert, they will not be issued repeat certificate install commands. Only the devices that require the cert will receive install commands.

**Certificate Installation**

After a command has installed the certificate on a device, the commands generated from this release should also update the system insights table for each device. 

For macOS devices there are no additional considerations, the generated command should be able update the device certificate table. 

For windows devices, there's an additional requirement to install a specific version of curl before we can use the agent generated certs. The version of curl.exe installed in windows 10 doesn't allow for .crt files to be passed in without converting the cert file to a .pem. Of the two compromises, installing an optional version of curl seemed to be the least egregious option. If the required version of curl is not installed, the installation command will skip over the step to update system insights with the latest certificate information. It could take up to an hour before the system insights table is updated and JumpCloud is made known that a system has a certificate installed.  

Screenshot 2024-06-03 at 3 19 57 PM

## How should this be tested?

- Have both a windows and macOS VM with each associated to a different JumpCloud user
- On the Windows VM, be sure to install "curl" with chocolatey using software management in the JumpCloud admin console. This is required to update the system insights tables after certificate installation.
- Have both of those users in a user group, associate that user group to some radius server. take note of the user Group ID 
- Download this branch and populate the user group ID in the `config.ps1` file
- Generate a root cert if you've not already done so
- Generate user certs for the two users 
- Distribute the user certs for the two users
- Wait for the certs commands to install
- Close and run the tool again both certs should be listed as already installed

## Screenshots

[CUT-4090]: https://jumpcloud.atlassian.net/browse/CUT-4090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gweinjc (approved):
Nice work! Great write up and everything is working as expected. This will be a huge benefit for customers!

kmaranionjc (approved):
Nice work here, no issues testing on my end.

PR #599, opened 2024-08-26
Cut 4249 parallel cert lookup

jworkmanjc:
## Issues
* [CUT-4249](https://jumpcloud.atlassian.net/browse/CUT-4249) - Get Certs in Parallel Threads

## What does this solve?

This release creates another hash data file to track which user certificates have been installed on devices. The previous version of the tool would check each user's sha1 cert value against the JumpCloud API in a loop. This release does the same thing but in parallel threads and during the Get-JCRGlobalVars function calls. When the tool first loads, the tool will check if the `certHash.json` file exists and if not populate it with data from the `users.json` file. 

When a user certificate is installed on a device, it should show up in the `certHash.json` file. The only difference in this release and the previous iteration of the tool is that the all the certs are validated and stored in some file until the cache is updated. This should reduce the time required to update data/ start the tool.

## How should this be tested?

Generate some certs for a set of users. Generate the commands to install the certs on a set of devices. With at least one of the devices (ideally several of the devices) on, run the commands. The certs should install. Update the global variables by pressing `5` on the main menu. The `certHash.json` file should be updated to reflect the certs that were applied to each device.

## Screenshots

Screenshot 2024-08-26 at 4 30 41 PM

[CUT-4249]: https://jumpcloud.atlassian.net/browse/CUT-4249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kmaranionjc (approved):
Awesome work here! Code looks good

PR #610, opened 2024-09-30
Initial commit of the Windows uninstall script

ecourtneyjc:
## Issues
* [<jira-id>](https://jumpcloud.atlassian.net/browse/SUP-1458) - <jira-title>

## What does this solve?
Gives support and customers a script to automatically remove broken Windows agent installs.

Since the GUID used in the registry to identify the current installation is random, discovery of the JC Agent GUID is done via recursive search through 'HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Products\' to obtain the ID used to remove the remaining keys.

## How should this be tested?
Install the JumpCloud Windows Agent
Confirm the existence of the following registry entries

- HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Products\<JumpCloud Agent GUID>
- HKCR:\Installer\Products\<JumpCloud Agent GUID>
- HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\<JC MSI Installation ID>
- HKLM:\Software\JumpCloud
- HKLM:\SOFTWARE\Classes\Installer\Products\<JumpCloud Agent GUID>

Run the UninstallWindowsAgent.ps1 script as an administrator.
Verify the above keys are removed
Verify the JumpCloud Agent service is no longer present.
Verify that the JumpCloud Agent folder is empty aside from:
- cpoidc
- JumpCloudCredentialProvider-X.XXX.X.dll

Reinstall Windows Agent using preferred method.
Confirm agent reinstall is successful.

### DIFF
--- a/scripts/windows/UninstallWindowsAgent.ps1
+++ b/scripts/windows/UninstallWindowsAgent.ps1
@@ -1,5 +1,5 @@
-# Function to check if Powershell is running as Administrator
-function Test-Administrator {
+ # Function to check if running as Administrator
+ function Test-Administrator {
     $currentUser = [Security.Principal.WindowsIdentity]::GetCurrent()
     $principal = New-Object Security.Principal.WindowsPrincipal($currentUser)
     return $principal.IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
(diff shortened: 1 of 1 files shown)


jworkmanjc (commented):
Hey I tested this just now. Two thoughts:

The script does tell me that the uninstall Key does not exist. And I believe it's due to the fact that the uninstall key looks something like this:

`Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{E14D51FF-380F-4964-9581-665DED7F58C1}`

Where curly braces encase the GUID key.

Screenshot 2024-10-07 at 9 32 42 AM

Though everything seems to be removed I think the messaging is confusing here. 

FWIW my registry keys were as follows:
`Computer\HKEY_CLASSES_ROOT\Installer\Products\FF15D41EF0834694591866D5DEF7851C

Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Installer\UserData\S-1-5-18\Products\FF15D41EF0834694591866D5DEF7851C
Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{E14D51FF-380F-4964-9581-665DED7F58C1}
Computer\HKEY_LOCAL_MACHINE\Software\JumpCloud
Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Classes\Installer\Products\FF15D41EF0834694591866D5DEF7851C`

I'm curious if you are seeing the same verbose messaging @ecourtneyjc?

scripts/windows/UninstallWindowsAgent.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+        foreach ($key in $keysToDelete) {
+            if (Test-Path $key) {
+                Remove-Item -Path $key -Recurse -Force -Verbose -ErrorAction SilentlyContinue
+            } else {
+                Write-Host "Registry key $key not found."
+            }
+        }
+
+        # Stopping and removing the service
```
kmaranionjc, line 74: Should Stop-Service be before removing reg keys? A running agent might mess with key removals

scripts/windows/UninstallWindowsAgent.ps1
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    return $jumpcloudGUIDs
+}
+
+# Function to remove registry keys, JC folder, and JC service
+function Remove-JumpCloud {
+    $guids = Find-JumpCloudGUID
+
+    foreach ($guid in $guids) {
```
kmaranionjc, line 56: I got GUID and MSI Install ID when I ran `Find-JumpCloudGUID`. Both should not be looped together.

scripts/windows/UninstallWindowsAgent.ps1
```diff
@@ -0,0 +1,95 @@
...
+            if (Test-Path $key) {
+                Remove-Item -Path $key -Recurse -Force -Verbose -ErrorAction SilentlyContinue
+            } else {
+                Write-Host "Registry key $key not found."
+            }
+        }
+
+        # Stopping and removing the service
+        $serviceName = "jumpcloud-agent"
+        if (Get-Service -Name $serviceName -ErrorAction SilentlyContinue) {
```
kmaranionjc, line 76: This was ran twice since I have two `$guids` values 
image

scripts/windows/UninstallWindowsAgent.ps1
```diff
@@ -0,0 +1,95 @@
...
+        $jumpcloudSoftwareKey = "HKLM:\Software\JumpCloud"
+        $jumpcloudClasses = "HKLM:\SOFTWARE\Classes\Installer\Products\$guid"
+
+        $keysToDelete = @($installerKey, $classesKey, $uninstallKey, $jumpcloudSoftwareKey)
+
+        foreach ($key in $keysToDelete) {
+            if (Test-Path $key) {
+                Remove-Item -Path $key -Recurse -Force -Verbose -ErrorAction SilentlyContinue
+            } else {
+                Write-Host "Registry key $key not found."
```
kmaranionjc, line 70: Another instance of MSI installation ID on the `$guids` variable.
image

kmaranionjc (changes requested):
Hi Ed, thanks for the work here. I added some review points. Couple things that pop up from my tests is: 
1. Find-JumpCloudGUID gets both JCAGent GUID and MSI installation ID. It should be separated since it get's called and looped through `Remove-JumpCloud` function which results on some errors.

scripts/windows/UninstallWindowsAgent.ps1
```diff
@@ -0,0 +1,95 @@
...
+if (-not (Test-Administrator)) {
+    Write-Warning "This script needs to be run as an administrator."
+    exit
+}
+
+# Create HKCR Mapping
+New-PSDrive -Name "HKCR" -PSProvider Registry -Root "HKEY_CLASSES_ROOT" -ErrorAction SilentlyContinue
+
+# Function to recursively search for DisplayName in the registry
+function Find-JumpCloudGUID {
```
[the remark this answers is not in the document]
ecourtneyjc: I don't believe I've ever seen multiple agent installation GUIDS. Successful agent upgrades should remove any previous GUIDs as part of the upgrade process. I'll add a separate function for the Uninstall GUID to the new script.
Thanks!

scripts/windows/UninstallWindowsAgent.ps1
ecourtneyjc: No, the service runs independent from the reg keys. We'll stop and remove the service before removing the JC folder, but the keys can be removed at any point.

scripts/windows/UninstallWindowsAgent.ps1
ecourtneyjc: Do you know how you ended up with multiple $guid values? I'll test for this scenario and make the necessary changes, but I don't think I've come across that issue in the wild.

scripts/windows/UninstallWindowsAgent.ps1
ecourtneyjc: Changed the code regarding keys not found and added variable for uninstall GUID.

scripts/windows/UninstallWindowsAgent.ps1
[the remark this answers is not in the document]
ecourtneyjc: Added variable for the handling the Uninstall GUID.

jworkmanjc (approved):
Thanks for the changes, seems to be working for me. 

Small issue with re-installing while using PWSH 7.x versions, that should be addressed in https://github.com/TheJumpCloud/support/pull/609. Otherwise everything worked as described.

kmaranionjc (approved):
Awesome work here Ed!

PR #611, opened 2024-09-30
CUT-4281: Set-JCUser Null Manager fix

gweinjc:
## Issues
* [CUT-4281](https://jumpcloud.atlassian.net/browse/CUT-4281) - Set-JCUser Null Manager fix

## What does this solve?
Due to the way that PowerShell treats string datatypes, you are unable to truly set a string parameter to null. The manager field requires that the manager field be set to strictly null, an empty string will not work.

In order to solve for this, we simply need to remove the string datatype on the manager parameter. This will default that parameter to accepting 'object' input, which will work in this scenario

N/A

## How should this be tested?
1. Import the module
2. Attempt to set a user's manager to $null
```powershell
Set-JCUser -id <UserID> -manager $null
```
4. This should now remove any existing values in the user's manager field

[CUT-4281]: https://jumpcloud.atlassian.net/browse/CUT-4281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kmaranionjc (approved):
Test worked when setting manager to $null. Great job! Looks good to me

jworkmanjc (approved):
Tested with Set-JCUser and also validated that the Update-JCUserFromCSV functions were not affected by this change. That appears to be the case. 

Looks good to me!

PR #630, opened 2024-12-11
CUT-4394 - Update Device from CSV Function

gweinjc:
## Issues
* [CUT-4394](https://jumpcloud.atlassian.net/browse/CUT-4394) - Update Device from CSV Function
* [CUT-4490](https://jumpcloud.atlassian.net/browse/CUT-4490)- Import-JCUserFromCSV - bug adding more than 10 custom attributes

## What does this solve?
Introduces the `Update-JCDeviceFromCSV` and `New-JCDeviceUpdateTemplate` functions to the PowerShell Module

- **New-JCDeviceUpdateTemplate**: Generates a CSV template that is used for `Update-JCDeviceFromCSV`
- **Update-JCDeviceFromCSV:** This function will allow users to bulk update devices in their org via a CSV that is generated from the `New-JCDeviceUpdateTemplate` function
- There was also a fix for a bug with **Get-JCSystemApp** returning an error when searching MacOS devices for apps that are less than 4 characters long and do not have the .app extension in the search

N/A

### New-JCDeviceUpdateTemplate
```powershell
# This will display a GUI for the end user to interact with to pick and choose which fields they would like to be present in the CSV thats generated
New-JCDeviceUpdateTemplate
```
```powershell
# This will forcefully generate the CSV
New-JCDeviceUpdateTemplate -force
```

### Update-JCDeviceFromCSV
1. Edit the generated CSV from the previous tests with random filler

```powershell
# This will prompt with a GUI confirming if youd like to make changes with a Y or N prompt
# It will then iterate through each device and update their respective properties and output the updated device info
Update-JCDeviceFromCSV -CSVFilePath '/Path/To/file.csv'
```
```powershell
# This will forcefully update all the devices without prompts for the user and outputs the updated device info
# It will then iterate through each device and update their respective properties and output the updated device info
Update-JCDeviceFromCSV -CSVFilePath '/Path/To/file.csv'
```

### Get-JCSystemApp
```powershell
# This should no longer error when searching for an app on a MacOS device that is shorter than 4 letters without the .app extension
Get-JCSystemApp -ID <MacOS ID> -Name VLC
Get-JCSystemApp -Name VLC
```

[CUT-4394]: https://jumpcloud.atlassian.net/browse/CUT-4394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CUT-4490]: https://jumpcloud.atlassian.net/browse/CUT-4490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


PowerShell/JumpCloud Module/Public/Utilities/CSV_Import/New-JCDeviceUpdateTemplate.ps1
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            if ($ConfirmDeviceDescription -eq 'Y') {
+                $CSV.add('description', $null)
+            }
+
+            elseif ($ConfirmDeviceDescription -eq 'N') {
+            }
+
+            Write-Host "`nWould you like to update device hostnames?"
```
jworkmanjc, line 134: I would like to update hostnames but what will this do?

PowerShell/JumpCloud Module/Public/Utilities/CSV_Import/Update-JCDeviceFromCSV.ps1
```diff
@@ -0,0 +1,110 @@
+Function Update-JCDeviceFromCSV () {
```
jworkmanjc, line 1: Just an initial thought, this works just fine and I think like intended with the exception of this: when you go ahead and create a template with all the headers and submit update for just the device description, the function will fail.
Screenshot 2024-12-17 at 12 17 51 PM

Would it be possible to check if there's a null value in the CSV to just ignore the item in each row?

PowerShell/JumpCloud Module/Tests/Public/Systems/Get-JCSystemApp.Tests.ps1
```diff
@@ -229,4 +229,9 @@ Describe -Tag:('JCSystemApp') 'Get-JCSystemApp' {
         Get-JCSystemApp -SystemID $mac._id -name "Chess.App" | Should -Not -BeNullOrEmpty
         Get-JCSystemApp -SystemID $mac._id -name "Chess" | Should -Not -BeNullOrEmpty
     }
+    It "Tests to make sure an error is not thrown if an app is less than 4 characters long" {
```
jworkmanjc, line 232: This works, thank you!

PowerShell/JumpCloud Module/Tests/Public/Utilities/CSV_Import/Import-JCUsersFromCSV.Tests.ps1
```diff
@@ -636,6 +636,47 @@ Describe -Tag:('JCUsersFromCSV') "Import-JCUsersFromCSV 2.14.2" {
         }
     }
 }
+Describe -Tag:('JCUsersFromCSV') "Import-JCUsersFromCSV 2.15.1" {
```
jworkmanjc, line 639: Already tested this but noting that the tests have been validated!

PowerShell/JumpCloud Module/Tests/Public/Utilities/CSV_Import/Update-JCDeviceFromCSV.Tests.ps1
```diff
@@ -0,0 +1,28 @@
+Describe -Tag:('JCDeviceFromCSV') 'Update-JCDeviceFromCSV' {
```
jworkmanjc, line 1: Regarding the comment above the only other test I can think of here would be updating a system with a NULL value in some of the fields like systemInsights. 

I DO think that customers will download the CSV, set everything up skip over columns (like you can do in Update-JCUsersFromCSV) and try to use this tool.

jworkmanjc (changes requested):
Let's align on whether or not we need account for null items in the CSV, otherwise this looks quite complete.

PowerShell/JumpCloud Module/Public/Utilities/CSV_Import/Update-JCDeviceFromCSV.ps1
kmaranionjc: Got the same issue here as well. Ignoring a blank field/row would be ideal

kmaranionjc (changes requested):
Commands runs good, added comment change for the docs and also issue with null/blank values on CSV

jworkmanjc (approved):

kmaranionjc (approved):

PR #631, opened 2024-12-12
Cut-4468-cascade-manager-remove-jc-user

kmaranionjc:
## Issues
* [Cut-4468](https://jumpcloud.atlassian.net/browse/Cut-4468) - Cut-4468-cascade-manager-remove-jc-user

## What does this solve?
Adds functionality for cascading manager with Remove-JCUser
n/a
## How should this be tested?

`Remove-JCUser -UserId id`

- Delete a user/manager. If the user is a manager, there will be prompts to cascade the manager of the users
image
- Delete manager1 that is managed by manager2. Instead of cascading the users manager to manager2. Enter a different user id to manage the managed users.
image
- Delete manager1 that is managed by manager2. Cascade the manager2 to managed users
image

`Remove-JCUser -UserId id -force`

- Remove a user(manager) with force param. This should set its managed users manager to null

`Remove-JCUser -UserId id -force -CascadeManager user -CascadeManagerUser testUser`

- Should throw an error. CascadeManager and force is not allowed

`Remove-JCUser -UserId id -CascadeManager User -CascadeManagerUser Username/Id`
image
image

- Remove a user (manager). Managed users manager should be the given Cascade Manager

`Remove-JCUser -UserId id -CascadeManager auto`

- Setup `manager1` that is managed by `manager2`. This command should automatically cascade the manager of the managed users  of manager1 to manager2 and remove/delete manager1

`Remove-JCUser -UserId id -CascadeManager null`

- Remove a user/manager. It's managed users should have no manager after removal.


PowerShell/JumpCloud Module/Docs/Remove-JCUser.md
```diff
@@ -40,6 +50,28 @@ PS C:\> Remove-JCUser cclemons -Force
 ```
 
 Removes the JumpCloud User with Username 'cclemons' using the -Force Parameter. A warning message will not be presented to confirm this operation.
+If the user is a manager, this parameter will clear the manager field for the managed users by this manager.
+
+### Example 3
```
jworkmanjc, line 55: We need to note here that this is the same as running 

`Remove-JCUser cclemons -force`

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -26,39 +26,64 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
         [Switch]
         $ByID,
-
+        # Do not use $CascadeManager if $force is used
         [Parameter(HelpMessage = 'A SwitchParameter which suppresses the warning message when removing a JumpCloud User.')]
         [Switch]
-        $force
+        $force,
+        [Parameter(HelpMessage = 'A SwitchParameter for Cascading the manager of the user to the users managed by the user. NULL, AUTOMATIC (bubble up), ID (prompt for manager ID)')]
+        [ValidateSet('NULL', 'Auto', 'ID')]
```
jworkmanjc, line 34:
```suggestion
        [ValidateSet('Null', 'Automatic', 'Id')]
```

Let's talk about this as a team but I want these to be same case. Should we keep these param sets as is? Do they make sense for the customer?

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -26,39 +26,64 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+        # Validate dynamic parameter
+        if ($CascadeManager -eq 'ID') {
+            $CascadeManagerId = $PSBoundParameters['CascadeManagerId']
+            # Validate if the Id is a JC User from the $UserHash
+            if ($UserHash.ContainsKey($CascadeManagerId)) {
+                Write-Debug "CascadeManagerId is a valid JumpCloud User"
+            } else {
+                Write-Error "User does not exist. Please enter a valid UserID."
+                # Exit the script
+                Exit
```
jworkmanjc, line 84: Why would we exit here, this doesn't make sense? 

Try to run your code 
`Remove-JCUser -username "jeff.fisher" -CascadeManager id -CascadeManagerId bob` with a managerID that doesn't exist. Tell me if exit is the right choice here.

In addition, we should have a test for this.

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -26,39 +26,64 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
         }
 
         $deletedArray = @()
-
-        if ($PSCmdlet.ParameterSetName -eq 'Username' ) {
-            $UserHash = Get-DynamicHash -Object User -returnProperties username
-            $UserCount = ($UserHash).Count
-            Write-Debug "Populated UserHash with $UserCount users"
+        # If $cascadeManager and $force are used, throw an error
+        if ($CascadeManager -and $force) {
```
jworkmanjc, line 70: It would be my preference to keep checking the `PSCmdlet.parameterSetName` in this section to:

1) remain consistent with how we are validating the other parameters
2) prevent accidental variable assignment triggering this code block.

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+        } elseif ($PSCmdlet.ParameterSetName -eq 'UserID') {
+            # Validate if the Id is a JC User from the $UserHash
+            if ($UserHash.ContainsKey($UserID)) {
+                Write-Debug "UserID is a valid JumpCloud User"
+                $UserId = $UserHash.GetEnumerator().Where({ $_.Name -contains ($UserID) }).Name
+                $Username = $UserHash.GetEnumerator().Where({ $_.Name -contains ($UserID) }) | Select-Object -ExpandProperty Value | Select-Object -ExpandProperty username
+            } else {
+                Write-Error "User does not exist. Please enter a valid UserID."
+                # Exit the script
+                Exit
```
jworkmanjc, line 104: Again, test this code in your own environment and report back if `Exit` is appropriate

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
             } else {
                 Throw "Username does not exist. Run 'Get-JCUser | Select-Object username' to see a list of all your JumpCloud users."
             }
+        } elseif ($PSCmdlet.ParameterSetName -eq 'UserID') {
+            # Validate if the Id is a JC User from the $UserHash
+            if ($UserHash.ContainsKey($UserID)) {
+                Write-Debug "UserID is a valid JumpCloud User"
```
jworkmanjc, line 98:
```suggestion
                Write-Debug "UserID is a valid JumpCloud User"
```

which UserID? this should return the variable used to trigger the debug message

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
-        if ($PSCmdlet.ParameterSetName -eq 'UserID' ) {
-            $Username = $UserID
+        # Check if the user is a manager
+        if ($UserHash.Values.manager -contains ($UserID)) {
+            $isManager = $true
+            # Count the number of users the manager is managing
+            # $managerCount = ($UserHash.Values.manager -eq $UserID).Count
+            # Save each user the manager is managing in a list
+            $managedUsers = $UserHash.GetEnumerator().Where({ $_.Value.manager -eq $UserID }).Name
+            Write-Debug "Manager is managing $managedUsers users"
```
jworkmanjc, line 114: Let's print the variable username/ id here instead of just calling them "manager" in debug

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+            $hasManagerId = Get-JcSdkUser -Id $UserID | Select-Object -ExpandProperty manager
+            Write-Debug "Manager is managed by $hasManagerId"
```
jworkmanjc, line 116: Let's print the variable username/ id here instead of just calling them "manager" in debug

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+        } else {
+            $isManager = $false
+            Write-Debug "User is not a manager"
```
jworkmanjc, line 119: User should be ID or username in debug

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
-                $delete = Invoke-RestMethod -Method Delete -Uri $URI -Headers $hdrs -UserAgent:(Get-JCUserAgent)
-                $Status = 'Deleted'
-            } catch {
-                $Status = $_.ErrorDetails
-            }
-
-            $FormattedResults = [PSCustomObject]@{
-                'User'    = $Username
-                'Results' = $Status
+            if ($CascadeManager -and $isManager) {
```
jworkmanjc, line 123: Let's use $PSBoundParameters['CascadeManager'] to keep things consistent and avoid accidental variable assignment triggering this code section

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+                        $managerUsername = $UserHash.GetEnumerator().Where({ $_.Name -contains ($hasManagerId) }) | Select-Object -ExpandProperty Value | Select-Object -ExpandProperty username
+                        $cascade_manager = Read-Host "User $($Username) is managed by manager: $($managerUsername). Do you want to reassign the managed users to the manager: $($managerUsername)? (Y/N)"
+                        if ($cascade_manager -eq 'Y') {
+                            $newManagerId = $hasManagerId
+                            $Status = Delete-JCUser -Id $UserID -managerId $newManagerId -Headers $hdrs -UserHash $UserHash
+                        } elseif ($cascade_manager -eq 'N') {
+                            $newManagerId = Read-Host "Enter the UserID of the new manager"
+                            # Validate if the Id is a JC User
+                            try {
+                                $validateUser = Get-JcSdkUser -Id $newManagerId
```
jworkmanjc, line 153: Every user is already in the hash on line 74, we should avoid making another unnecessary call to validate the user here

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+                                Write-Error "User does not exist. Please enter a valid UserID."
+                                # Exit the script
+                                Exit
+                            }
+                        }
+                    } else {
+                        $newManagerId = Read-Host "Enter the UserID of the new manager"
+                        # Validate if the Id is a JC User
+                        try {
+                            $validateUser = Get-JcSdkUser -Id $newManagerId
```
jworkmanjc, line 166: Every user is already in the hash on line 74, we should avoid making another unnecessary call to validate the user here

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+                            Write-Debug "User $newManagerId is a valid JumpCloud User"
+                            $Status = Delete-JCUser -Id $UserID -managerId $newManagerId -Headers $hdrs -UserHash $UserHash
+                        } catch {
+                            Write-Error "User does not exist. Please enter a valid UserID."
+                            # Exit the script
+                            Exit
```
jworkmanjc, line 172: Is `Exit` appropriate here?

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+                        }
+                    }
+                } elseif ($cascade_manager -eq 'N') {
+                    $Status = Delete-JCUser -Id $UserID -managerId $null -Headers $hdrs -UserHash $UserHash
+                } else {
+                    Write-Error "Please enter Y or N"
+                    Exit
```
jworkmanjc, line 179: Review use of `Exit`

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
+                        if ($cascade_manager -eq 'Y') {
+                            $newManagerId = $hasManagerId
+                            $Status = Delete-JCUser -Id $UserID -managerId $newManagerId -Headers $hdrs -UserHash $UserHash
+                        } elseif ($cascade_manager -eq 'N') {
+                            $newManagerId = Read-Host "Enter the UserID of the new manager"
+                            # Validate if the Id is a JC User
+                            try {
+                                $validateUser = Get-JcSdkUser -Id $newManagerId
+                                Write-Debug "User $newManagerId is a valid JumpCloud User"
+                                $Status = Delete-JCUser -Id $UserID -managerId $newManagerId -Headers $hdrs -UserHash $UserHash
```
jworkmanjc, line 155: I understand why you are running this as a private function `Delete-JCUser` should have it's name changed IMO. 

I'd like our functions to use [PS Approved Verbs](https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands?view=powershell-7.4).

From that link:
"Do not use a synonym of an approved verb. For example, always use `Remove`, never use `Delete` or `Eliminate`."

Which begs the question, why bubble this out into a private function if the name was going to be `Remove-JCUser` anyways. I feel like a better choice might be this:

You have a need to essentially change the URL depending on the actions necessary here. Here are the possible actions and feel free to correct me if I'm wrong:

Two possible options:
1. A user is being deleted and they are not managing other users
2. A user is being deleted and they are managing other users

If they are a manager of other users
1. Set the users subordinates user's manager to be `Null`
2. Set the users subordinates manager to be the `manager` of the user being deleted
3. Set the users subordinates manager to be some new user

This set of code is confusing to me because we have read hosts here, we have read hosts in `Delete-JCUser`. So I'm wondering what the point of the abstraction does here?

If all we need to do is determine if the user being deleted is a manager or not. AND the action we should take IF they are manager. I think this could be simplified into one of two ways:

1. Bring the code back into `Remove-JCUser` and within your tree of conditions, just return a URL string instead of passing the code to `Delete-JCUser`
2. If there's a need to break up the code, create a new function like `Confirm-JCUserForRemoval` that validates that the customers actions will not return some error and return a URL. 

In this way I think the code could be a bit easier to read but let's discuss.

PowerShell/JumpCloud Module/Private/Users/Delete-JcUser.ps1
```diff
@@ -0,0 +1,67 @@
...
+    }
+    process {
+        try {
+            if ($managerId) {
+                $URI = "$JCUrlBasePath/api/systemusers/$($Id)?cascade_manager=$($managerId)"
+                # Get the manager's username from $Userhash
+                $managerUsername = $UserHash.GetEnumerator().Where({ $_.Name -contains ($managerId) }) | Select-Object -ExpandProperty Value | Select-Object -ExpandProperty username
+                Write-Host "Deleting user: $($Username) and cascading managed users to manager: $($ManagerUsername)" -ForegroundColor Yellow
+                $prompt = Read-Host "Are you sure you wish to delete the user: $($Username)? (Y/N)"
+                while ($prompt -ne 'Y' -and $prompt -ne 'N') {
```
jworkmanjc, line 23: Just per the other comments, I'm confused why we have a read host in `Remove-JCUser` and are again prompting within the private function. This should be reworked IMO to only ask for input in one place and have actions depending on that behavior.

jworkmanjc (changes requested):
Let's chat about this as a team, I think there are a few changes that have to be made, there are a few changes that could be made to make this more readable.

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
gweinjc: Yeah I believe throw would be more appropriate in this scenario

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
gweinjc: I'm of the opinion that this should all be self contained in Remove-JCUser.

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
gweinjc: I'm slightly torn on this.

At first glance when testing I was like this doesn't make much sense but I understand why would do it this way as well. If we want tab completion we want to use ValidateSet, but we could potentially leverage ValidateScript and check for $null, Automatic/Auto and the pattern for a userID and then forego the dynamicparams

Just a thought

gweinjc (changes requested):
I believe we need to change the branch to be merged into CUT-4394_JCDeviceFromCSV instead of master if we intend to release both features

Also when testing, there isn't any feedback whether or not the manager was actually cascaded or not, some kind of output might be beneficial to customers showcasing that the new manager assignment took place

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
```diff
@@ -67,52 +92,106 @@ UserID has an Alias of _id. This means you can leverage the PowerShell pipeline
...
                 Throw "Username does not exist. Run 'Get-JCUser | Select-Object username' to see a list of all your JumpCloud users."
             }
+        } elseif ($PSCmdlet.ParameterSetName -eq 'UserID') {
+            # Validate if the Id is a JC User from the $UserHash
+            if ($UserHash.ContainsKey($UserID)) {
+                Write-Debug "UserID is a valid JumpCloud User"
+                $UserId = $UserHash.GetEnumerator().Where({ $_.Name -contains ($UserID) }).Name
+                $Username = $UserHash.GetEnumerator().Where({ $_.Name -contains ($UserID) }) | Select-Object -ExpandProperty Value | Select-Object -ExpandProperty username
+            } else {
+                Write-Error "User does not exist. Please enter a valid UserID."
```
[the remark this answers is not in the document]
kmaranionjc: $UserId is not needed in this call since we are already validating with the if statement. 
image

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
kmaranionjc: Done
image

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
kmaranionjc: Done: 
image

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
kmaranionjc: Done: image

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
kmaranionjc: It's consistent with our JCUser commands. When we input an invalid ID to Set-JCUser, an error message is thrown. 
image

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
kmaranionjc: Read-Host and abstraction wise, we are calling Read-Hosts with our other private functions so I thought it would be okay. The Read-Host calls in Delete-JCUser is only used to confirm the deletion of user, no IDs or usernames being prompted. Remove-JCUser on the other hand, we ask for `Manager ID` or `CascadeManagerId` and validating it in the same function before calling Delete-JCUser to invoke the deletion.

kmaranionjc: Naming convention wise, `Invoke-JCUserRemoval`?

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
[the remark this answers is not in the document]
jworkmanjc: In this case I'm asking your error messaging to be more specific — what user does not exist? The username of the user being deleted or the manager? 

Just noting that when we error we should be providing helpful messaging for customers, I'd argue we need to say what user (and specify the entered text ID or otherwise) does not exist. If I got this message while using the function, I'm not sure I'd know what user does not exist. Make sense?

PowerShell/JumpCloud Module/Public/Users/Remove-JCUser.ps1
kmaranionjc: Moved Invoke-restmethods and removed the private function

PowerShell/JumpCloud Module/Private/Users/Delete-JcUser.ps1
kmaranionjc: Moved the functionality to remove-JCUser

jworkmanjc (changes requested):
Just a few wording updates in the examples really, otherwise I think this is okay

PR #641, opened 2025-01-21
BAN 2695: MQTT troubleshooting scripts added to commands gallery

asinghjc:
## Issues
* [BAN-2695](https://jumpcloud.atlassian.net/browse/BAN-2695) - MQTT troubleshooting scripts added to commands gallery


gweinjc (approved):
Commands tested fine in my environment

jworkmanjc (approved):
The only problem I had running these scripts was on linux, netcat was not installed by default, I think we'll see that often if admins are asked to run this. You do mention that it's required so I'm okay to go for it and see if it's an issue. We can always add other messaging later.

PR #647, opened 2025-01-31
Tests for radius membership

jworkmanjc:
## Issues
* [CUT-4093](https://jumpcloud.atlassian.net/browse/CUT-4093) - Validate membership tests

## What does this solve?

- If large numbers of certificates are stored in the user certificates directory this update to the module will parse only the certificates that match usernames of the users in the defined radius group.
- BugFix for cases where a user was in the radius member group but their cert could not be identified if they were the first user in the group.
- Update tests to expand expected behavior

## How should this be tested?
Automated tests cover these changes

## Screenshots

[CUT-4093]: https://jumpcloud.atlassian.net/browse/CUT-4093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/automation/Radius/Functions/Private/CertDeployment/Get-CertInfo.ps1
+++ b/scripts/automation/Radius/Functions/Private/CertDeployment/Get-CertInfo.ps1
@@ -23,10 +23,17 @@ function Get-CertInfo {
         if ($UserCerts) {
             # Find all userCert paths
             if ($username) {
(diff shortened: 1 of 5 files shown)


kmaranionjc (approved):
Thanks for the bug fix! Looks good to me

gweinjc (approved):

PR #664, opened 2025-04-01
SUP-1514: Amended Device Cert capture

jwainwrightjc:
## Issues
* [SUP-1514](https://jumpcloud.atlassian.net/browse/SUP-1514) - Device Trust Certificate capture update

## What does this solve?
Currently, the function for listing / capturing Device Trust Cert within the script will only capture if and for the currently logged-in user on the device - Though in most cases this will typically not be an issue, if the log collection is run and a non-managed user is logged in, the cert info wouldn’t be checked for any possible managed users on the device, possibly then returning a false negative.

This change is to move the current logic for capturing the device cert info into the FOR loop against managed users.

N/A

## How should this be tested?
Run the script from JC commands with no user logged in and confirm the Device Cert info and file are written for the managed users on the device.

## Screenshots
Screenshot 2025-04-01 at 09 34 54
Screenshot 2025-04-01 at 09 36 06

[SUP-1514]: https://jumpcloud.atlassian.net/browse/SUP-1514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/macos/log_collection.sh
+++ b/scripts/macos/log_collection.sh
@@ -11,7 +11,7 @@ days=2           # number of days of OS logs to gather
 # do not edit below
 #######
 
-version=1.2.3
+version=1.2.4
 
 ## verify script is running as root.
 if [ $(/usr/bin/id -u) -ne 0 ]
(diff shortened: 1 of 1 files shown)


gweinjc (approved):

kmaranionjc (approved):
Tests worked on my end. Thanks for the change! 

image

PR #667, opened 2025-04-02
Cut 4639 update module functions config

jworkmanjc:
## Issues
* [CUT-4639](https://jumpcloud.atlassian.net/browse/CUT-4639) - Module Update Functions & Config Changes

## What does this solve?

This release is a precursor to releasing this version of the module as a proper PowerShell Module


kmaranionjc:
Add a test fore $settings variable for spaces:
image

scripts/automation/Radius/Functions/Private/Config/Test-JCRRadiusDirectory.ps1
gweinjc, line 1: We could potentially create the parent directory if it does not exist as well instead of just throwing an error. Will make the user experience a bit more seamless

image

scripts/automation/Radius/Functions/Private/Config/Set-JCRRadiusDirectory.ps1
gweinjc, line 1: Empty file?

scripts/automation/Radius/Functions/Private/CertGeneration/Test-JCRExtensionFile.ps1
gweinjc, line 1: Getting the following error when attempting to create a new Root CA even though the Set-JCRConfigFile never prompted to set headers

image
image

```powershell
Set-JCRConfigFile -certSubjectHeader @{
>> CountryCode        = "US"
>> StateCode          = "CO"
>> Locality           = "Boulder"
>> Organization       = "JC"
>> OrganizationalUnit = "Customer_Tools"
>> CommonName         = "JumpCloud.com"
>> }
```

Even after setting the headers and re-attempting to generate a Root CA I'm getting the same error
image

kmaranionjc (changes requested):
Added review comments. But looks good for the most part

gweinjc (changes requested):
Ran into a few issues when attempting to regenerate a Root CA.

The initial run worked fine and was able to generate a Root CA, UserCerts as well as deploy them to devices. However when regenerating a new Root CA, ran into a few issues that probably could have been avoided on the first run

I was never prompted to set any certHeaders on the initial run. A root CA was generated regardless as well as userCerts. On a subsequent run I attempted to regenerate a rootCA and recieved a bunch of errors regarding the headers.

I then attempted to set the headers with the Set-JCRConfig function and the errors are still present when attempting to generate a new RootCA. Issue occurs in a fresh terminal window as well

jworkmanjc:
> Add a test fore $settings variable for spaces: image

This should be resolved now. When someone "Sets" the config, it'll test the settings and throw if there's a space in the cert subject headers.

Screenshot 2025-06-23 at 8 29 20 AM

There are additional tests written for this exact case in the "Confirm-JCRConfig.Tests.ps1" file

kmaranionjc (approved):
Thanks for the work here!

PR #674, opened 2025-04-22
Added some enhancements to PWM scripts

oalobajijc:
## Issues
* [NB-8076](https://jumpcloud.atlassian.net/browse/NB-8076) - Update PWM without requiring Admin permissions

## What does this solve?
****NOTE:
The acceptance criteria for the ticket is the following:
1. Check machine whether its ARM or Intel and install the respective build.
2. Check if there is a previous installation in /Applications and then move to /user directory
3. Force quite the app before the installation and re-launch after

We only support ARM builds for mac currently, this has been added in a previous PR
Moving the app to the user directory is also already added
This PRs purpose:
1. Makes sure no instance of PWM is running before performing app install for both Mac & Windows
2. Deletes hanging/corrupt/stale installation file for Windows 
3. Adds the ability to launch-after-installation for Mac
## How should this be tested?
The script should be ran from the admin portal to a device connected with the JC Agent
## Screenshots
image

[NB-8076]: https://jumpcloud.atlassian.net/browse/NB-8076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gweinjc (dismissed):
Appears to be working on my test machine

gweinjc (approved):

PR #677, opened 2025-04-25
CUT-4711-Nuspec-Path-Bug-Fix

kmaranionjc:
## Issues
* [CUT-4711(https://jumpcloud.atlassian.net/browse/CUT-4711) - CUT-4711-Nuspec-Path-Bug-Fix

## What does this solve?
The latest release of our workflow runner, when transitioning from Linux to Windows for Nuspec builds, has introduced a bug. This bug results in an incorrect directory structure for the `Public`, `Private`, and `en-Us` folders, where extra path information is included ex. `Private\Private\Association` instead of `Private\Association`. The problem is caused by the use of Linux-style forward slashes (`/`) instead of Windows-style backslashes (`\`) in the directory paths. Correcting these path separators resolves the issue.

## How should this be tested?
1. Build nuspec: `. /Users/YOURUSER/Documents/GitHub/support/PowerShell/Deploy/BuildNuspecFromPsd1.ps1`
2. Open the built nuspec file using the Nuget Package Explorer
3. Directory should like this: 
image

### DIFF
--- a/PowerShell/Deploy/BuildNuspecFromPsd1.ps1
+++ b/PowerShell/Deploy/BuildNuspecFromPsd1.ps1
@@ -8,9 +8,9 @@ param (
 . "$PSScriptRoot/Get-Config.ps1"
 # $nuspecFiles = @{ src = 'en-Us/**;Private/**;Public/**;JumpCloud.psd1;JumpCloud.psm1;LICENSE'; }
 $nuspecFiles = @(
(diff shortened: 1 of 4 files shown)


jworkmanjc (approved):

tonybjc (approved):
Just a swap of back/forward slashes, looks good. Approved.

PR #687, opened 2025-06-17
HY-3451-script-Add-automation-for-assignment-of-Primary-User

twiemannjc:
## Issues
* [HY-3451](https://jumpcloud.atlassian.net/browse/HY-3451) - Add automation for assignment of Primary User

## What does this solve?
As an admin, I want to automatically Primary User on as many of my existing devices as possible, so that I can reduce the amount of manual assignments necessary

There are multiple scenarios where Primary User should NOT be assigned, either because the business logic would reject the assignment (ex: indirect bindings), or because we cannot "pick" the Primary User for the admin (ex: multiple users bound).

## How should this be tested?
Either create multiple systems, or run the script multiple times against a single system, to validate that all scenarios are covered (see https://jumpcloud.atlassian.net/browse/HY-3451 for more info)

## Screenshots
N/A

### DIFF
--- a/scripts/automation/assign_primary_user.ps1
+++ b/scripts/automation/assign_primary_user.ps1
@@ -1,0 +1,276 @@
+<#
+.DESCRIPTION
+    This script identifies JumpCloud managed systems that can have a Primary User assigned. 
+    It allows an administrator to provide a "ignore list" of usernames to exclude from consideration. 
+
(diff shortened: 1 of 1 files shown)


kmaranionjc (approved):
Looks good on my end: 
image
image
image

jworkmanjc (approved):

PR #730, opened 2026-03-27
Create Windows - Enable Quick Machine Recovery & Set Parameters for J…

ReedSavoryArcadiaFunds:
JumpCloud

Commands Gallery

Enable Windows Quick Machine Recovery & Set Parameters for JumpCloud

Since JumpCloud doesn't have a policy for this I've developed a bulletproof JumpCloud command to enable the new Windows Quick Machine Recovery functionality and set the optional parameters for the feature, and which works universally on all Windows 11 platforms (both x86/x64 and Windows On ARM as well).

---

> [!NOTE]
> **Medium Risk**
> Enables and configures Windows recovery behavior via `reagentc`, which modifies OS-level recovery settings and could impact device remediation behavior if misconfigured or run on unsupported systems.
> 
> **Overview**
> Adds a new JumpCloud PowerShell command that enables Windows 11 Quick Machine Recovery and applies recovery settings by writing a base64-encoded XML config to disk and invoking `reagentc /enable` plus `reagentc /setrecoverysettings` under a forced 64-bit PowerShell context.
> 
> Includes inline reference presets (XML + base64) for common remediation schedules, with the script defaulting to an every-12-hours auto-remediation check, and returns a `TRUE`/`FALSE` result with step exit codes for JumpCloud reporting.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit df97a7b447988da56822ec2b5fba4957632e180d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Enable Quick Machine Recovery & Set Parameters for JumpCloud
+++ b/PowerShell/JumpCloud Commands Gallery/Windows Commands/Windows - Enable Quick Machine Recovery & Set Parameters for JumpCloud
@@ -1,0 +1,112 @@
+## Enter your command within the @' '@
+## Use single quotes ' not double quotes " for all variables
+## Example multi-line command below
+##
+## Enable Windows Quick Recovery for Windows 11 in JumpCloud
+## Platform agostic - works with Windows 11 on all platform variants (x86/x64, Windows on ARM, etc.)
+## Free for use by all JumpCloud customers
+## Release date: March 10, 2026
+## Reed Savory, rsavory@arcadiafunds.com
+## Arcadia Funds, LLC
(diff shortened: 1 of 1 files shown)


jworkmanjc:
Closing this, I think for the average consumer including myself the base64 encoded scripts should all be string statements then base64 encoded within the script. Although I can decode these variable assignments, I wouldn't want them to be published in our command gallery in this way. A better way to do this would be to define a variable for the wait interval, create the XML string with that variable, and convert that payload to base64 if needed.

For those that are interested in this script, it'll still show up in the closed Pull Requests. If there's a need for this please feel free to submit a [feature request](https://jumpcloud.com/support/request-features) and I can get it routed, tested properly and published.

PR #742, opened 2026-05-20
Cut 5088 updating power shell module to handle o auth/service account tokens

shashisinghjc:
## Issues
* [CUT-5088](https://jumpcloud.atlassian.net/browse/CUT-5088) - Updating PowerShell Module to handle OAuth/Service Account tokens

## What does this solve?
This adds a new authentication method i.e. ClientSecret(OIDC). Earlier only API key was allowed.

NA

## How should this be tested?
Use Connect-JCOnline - Default is the API key authentication
Use `Set-JCSettingsFile -authPreferenceMethod ClientSecret` (This will switch to ClientSecret) and opt for the client ID and Client Secret

## Screenshots

[CUT-5088]: https://jumpcloud.atlassian.net/browse/CUT-5088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **High Risk**
> This changes authentication across the module, stores client secrets and tokens in environment/session variables, and many cmdlets still build local `x-api-key` headers—clientSecret mode may not work uniformly until those paths use `Get-JCAuthHeaders`.
> 
> **Overview**
> Adds **OAuth client credentials** (`clientSecret`) as an alternative to the existing **API key** flow, controlled by a new `authPreference` setting (`apiKey` | `clientSecret`) in `Config.json`.
> 
> New private helpers mint and cache bearer tokens (`New-JCBearerToken`), track expiry (`Get-JCAccessToken`), pick the active method (`Get-JCActiveAuthMethod`), and build request headers (`Get-JCAuthHeaders`, with automatic refresh). **`Connect-JCOnline`** gains `-JumpCloudClientId` / `-JumpCloudClientSecret` and branches connect, org setup, and validation by auth mode. **`Invoke-JCApi`** and **`Get-JCResults`** now use centralized headers instead of hard-coded `x-api-key`.
> 
> Many public cmdlets replace direct `$JCAPIKEY` checks with **`Test-JCConnection`** before calling `Connect-JCOnline`. **`Update-JCSettingsFile`** rebuilds config from current defaults on merge so upgrades pick up new settings (including `authPreference`). Policy template tab completion loads via **`Invoke-JCApi`** so it works under bearer auth (CUT-5088).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbe4a0be4e46d65938b82ce5a8ad26e4f038559f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jworkmanjc (commented):
Okay I did add one change to support getting some data with ClientID/ClientSecret. I suspect we'll need to add more to clear this work up. 

One question I had was if it was really required to get a new token on each request. I'm running `Get-JCUser -username "someUser" -debug` to test the new flow and watching the flow from `Connect-JCOnline` -> `New-JCBearerToken`. We should be making use of the Token "Expires_in" value which I believe is in seconds so we don't have to do another request to auth every time we hit "Connect-JCOnline" Not sure if it's best to just set it as en Environment Variable or set it in the config. But we shouldn't be generating a new token each time I do a request for data. I think we can do something like this:

First ClientID/Secret request, Get-CurrentDate, get authToken, look at `expires_in` value and record (Current-Date + add expires_in(in seconds) as some env variable. Next time we do a request check if Current date -gt that Env variable and perform auth again if necessary.

What do you think about that @shashisinghjc? We'd have to do the same thing for OIDC auth right?

shashisinghjc:
hi @jworkmanjc , here's the flow. New-Bearer token is only called when the token is missing or expired.

Every API request flow:

Invoke-JCApi → calls Get-JCAuthHeaders
Get-JCAuthHeaders → calls Get-JCAccessToken
Get-JCAccessToken checks token validity:
Token is valid → Returns cached $env:JCAccessToken (no API call)
Token expired/missing → Calls New-JCBearerToken to fetch new token

New token is Stored in $env:JCAccessToken and $global:JCAccessToken. 
Expiration: Tracked in $env:JCAccessTokenExpiresAt

lucasmendes-jc (dismissed):

PowerShell/JumpCloud Module/Public/Authentication/Connect-JCOnline.ps1
```diff
@@ -97,81 +142,89 @@ function Connect-JCOnline () {
...
+                    $global:JCClientId = $env:JCClientId
+                }
+                if (-not [System.String]::IsNullOrEmpty($JumpCloudClientSecret)) {
+                    $env:JCClientSecret = $JumpCloudClientSecret
+                    $global:JCClientSecret = $env:JCClientSecret
+                }
+                if ([System.String]::IsNullOrEmpty($env:JCClientId) -or [System.String]::IsNullOrEmpty($env:JCClientSecret)) {
+                    throw 'ClientId and ClientSecret are required when authPreference is set to clientSecret. Pass -JumpCloudClientId and -JumpCloudClientSecret or run Set-JCSettingsFile -authPreferenceMethod apiKey to switch modes.'
+                }
+                $tokenInfo = New-JCBearerToken -ClientId $env:JCClientId -ClientSecret $env:JCClientSecret
```
jworkmanjc: Okay this is where we are getting a token on each call. 

Just run `Get-JCUser -username "some.user"` to validate, you can see, it'll generate a new token. One thing which may not make a log of sense but is currently in the module is the fact that `Connect-JCOnline` is called in each public function. That way if some user wants to run one of our functions, the tooling will prompt for authentication if it's not set. That means that each time you run a public function, a new bearer token is generated.

Image

jworkmanjc: One thing to do here would be to call `Get-JCAccessToken`, which IMO does more than it's supposed to. 

I know if you read my code, I might not be the best at practicing what I'm preaching but the "Get-" functions should only "Get". "New-" functions should only create. 

Here we should do a sequence like this:
Get-JCAccessToken - return accesstoken, validity date
   If exists, validate the expiration
   If missing, call New-JCBearerToken

I think that'll solve the issue moving forward

PowerShell/JumpCloud Module/Private/Authentication/Get-JCAccessToken.ps1
```diff
@@ -0,0 +1,23 @@
...
+
+        # 30s skew avoids the race where a token is valid at check-time but expires mid-request
+        $isExpired = (Get-Date).AddSeconds(30) -ge $expiresAt
+
+        if ([System.String]::IsNullOrEmpty($env:JCAccessToken) -or $isExpired) {
+            if ([System.String]::IsNullOrEmpty($env:JCClientId) -or [System.String]::IsNullOrEmpty($env:JCClientSecret)) {
+                throw 'Access token is missing or expired and ClientId/ClientSecret are not available. Run Connect-JCOnline to re-authenticate.'
+            }
+            Write-Verbose 'Access token missing or expired; requesting a new bearer token.'
+            New-JCBearerToken -ClientId $env:JCClientId -ClientSecret $env:JCClientSecret | Out-Null
```
jworkmanjc: In my opinion we should just be returning the token and expiration date here, not calling a new token, that would keep the private function more in line with it's intended function "GET"

PowerShell/JumpCloud Module/Private/settings/New-JCSettingsFile.ps1
```diff
@@ -16,17 +16,20 @@ function New-JCSettingsFile {
...
                 'Eligible'   = @{value = Get-JCParallelValidation; write = $false; copy = $true }
                 'Override'   = @{value = $false; write = $true; copy = $true }
                 'Calculated' = @{value = $false; write = $false; copy = $true }
             }
-            'JCEnvironment' = @{
-                'Location' = @{value = $JCEnvironmentValue; write = $true; copy = $true; validateSet = 'STANDARD STAGING EU IN' }
+            'JCEnvironment'  = @{
+                'Location' = @{value = $JCEnvironmentValue; write = $true; copy = $true }
+            }
+            'authPreference' = @{
```
jworkmanjc, line 31: I need to create a new card here, but the code to update/ set the validate set does not update when new versions of the module are released. This is out of scope for this PR but I'm marking it for context/ secondary reviewers

PowerShell/JumpCloud Module/Public/Authentication/Connect-JCOnline.ps1
```diff
@@ -240,21 +293,26 @@ function Connect-JCOnline () {
                 Write-Verbose "Error: Unable to set module authentication"
             }
             # set Argument Completer(s) which require authentication
-            $templates = Get-JcSdkPolicyTemplate
-            $global:TemplateNameList = New-Object System.Collections.ArrayList
-            foreach ($template in $templates) {
-                $templateHashObject = [PSCustomObject]@{
-                    Name = ("$($template.osmetafamily) $($template.displayname)").Replace(' ', '_')
-                    Id   = $template.Id
+            # TODO: CUT-5088 — JumpCloud.SDK cmdlets require a mandatory -ApiKey, so they can't yet
```
jworkmanjc: Yep, we need to address this. The argument complete could be moved, I'm not opposed to this.

PowerShell/JumpCloud Module/Public/Authentication/Get-JCAuthStatus.ps1
jworkmanjc, line 1: This function isn't used anywhere in the module, but I feel like it does what I want the "get-JCAccessToken" function to do in terms of returning valuable info about the token. 

Was there a plan with this function?

jworkmanjc (changes requested):
I did find that issue with `Connect-JCOnline` @shashisinghjc, we can chat about this tomorrow. I'll plan to join work early my time, would be good to sync up since I haven't gotten a chance to talk with you much this week. 

Take a look at the requested changes/ thoughts and let me know what you think. I really apologize it's taken me this long to review.

PowerShell/JumpCloud Module/Public/Authentication/Get-JCAuthStatus.ps1
shashisinghjc: Well I added this function intentionally as part of tests. To run and see the current auth method and configurations during debugging.

PowerShell/JumpCloud Module/Private/Authentication/Get-JCAccessToken.ps1
shashisinghjc: this is taken care

lucasmendes-jc:
As [security said](https://jumpcloud.atlassian.net/browse/SEC-10446?focusedCommentId=757660) that needs to be implemented as masked input: `the client secret must have a masked entry path`.

PR #758, opened 2026-09-17
CUT-5273: Add Set-JCWorkflow cmdlet and integration tests

junioralmeida-82:
## Issues
* [CUT-5273](https://jumpcloud.atlassian.net/browse/CUT-5273) - Set-JCWorkflow

## What does this solve?
Implements the `Set-JCWorkflow` cmdlet to update existing workflows via the `PUT /api/v2/workflows/{id}` endpoint. Includes Pester integration tests covering successful updates and error handling.

The `PUT` endpoint requires the full payload structure (including `dsl` and `execution_role_id`), which are handled during parameter mapping.

## How should this be tested?
1. Automated CI/CD execution via the `PowerShell Module` label.
2. Local Pester test execution:
   `Invoke-Pester -Path "./PowerShell/JumpCloud Module/Tests/Public/Workflows/Set-JCWorkflow.Tests.ps1"` (2/2 passing).

## Screenshots
N/A

[CUT-5273]: https://jumpcloud.atlassian.net/browse/CUT-5273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


edgar-lins (changes requested):
I pulled the branch and tested against a sandbox org before reviewing. One blocking issue, plus a couple of follow-ups.

**Partial updates fail — the primary use case is broken**

The .SYNOPSIS example Set-JCWorkflow -Id '...' -Name "Updated Workflow Name" doesn't work. Running it against the sandbox returns:

`{ "code": 602, "message": "dsl in body is required" }`

The endpoint isn't a partial `PATCH` — it requires the full payload, so any call that omits `dsl` (and likely `execution_role_id`) is rejected outright. Good news: because the whole request is refused, nothing gets corrupted — I confirmed `dsl` and the name were untouched afterward. Bad news: the cmdlet only works if you pass `-Dsl` on every call, which is why the current success test passes (it sends `Name` + `ExecutionRoleId` + `Dsl` together and never exercises the partial path).

Suggested fix — read-modify-write at the top of process to build the full body, filling unspecified fields from the current workflow:

```
$current = Get-JCWorkflow -Id $Id
$body = @{
    name              = if ($PSBoundParameters.ContainsKey('Name'))            { $Name }            else { $current.name }
    description       = if ($PSBoundParameters.ContainsKey('Description'))     { $Description }     else { $current.description }
    execution_role_id = if ($PSBoundParameters.ContainsKey('ExecutionRoleId')) { $ExecutionRoleId } else { $current.execution_role_id }
    dsl               = if ($PSBoundParameters.ContainsKey('Dsl'))            { $Dsl }             else { $current.dsl }
}
if ($PSBoundParameters.ContainsKey('Status')) { $body.Add('status', $Status) }
```
Worth double-checking that the `dsl` object returned by `Get-JCWorkflow` round-trips cleanly back through `ConvertTo-Json` into the shape the PUT expects.

**Test gap — add a partial-update case**
The AC asks for a "throw when parameters are invalid" case, but the negative test uses an invalid ID (a 404), not invalid params — and `Should -Throw` with no `-ExpectedMessage` passes on any error. More importantly, we're missing the test that actually covers the bug above:

```
It "Updates only the name and preserves dsl (partial update)" {
    $newName = "Pester_Partial_$(Get-Random)"
    $result  = Set-JCWorkflow -Id $script:testWorkflow.id -Name $newName
    $result.name | Should -Be $newName

    $after = Get-JCWorkflow -Id $script:testWorkflow.id
    $after.name | Should -Be $newName
    $after.dsl  | Should -Not -BeNullOrEmpty
}
```
This fails today (602) and passes once the read-modify-write is in — it locks the regression.

**Minor**

`catch { throw $_ }` is a no-op; either drop it or add context (`"Failed to update workflow $Id: $_"`).
`ValueFromPipeline = $true` on a `[String] $Id` will bind a piped object via `.ToString() `rather than by `.id`; consider `ValueFromPipelineByPropertyName` only.
No guard for an empty body — `Set-JCWorkflow -Id X` alone sends `{}`; a friendly `throw` when no updatable field is passed would help.
`Set-` cmdlet that mutates state — worth adding `SupportsShouldProcess` if the other `Set-JC*` cmdlets support -`WhatIf`/`-Confirm`, for consistency.

**What's solid**

Success test does a real round-trip assertion (`$after.name | Should -Be`), not just "didn't error."
`AfterAll` cleanup + `Get-Random` naming is good test hygiene.
The `WorkflowRole` / `role_id` aliases on `ExecutionRoleId` cleanly resolve the "name tbd" from the ticket.

PowerShell/JumpCloud Module/Tests/Public/Workflows/Set-JCWorkflow.Tests.ps1
```diff
@@ -0,0 +1,55 @@
...
+        $updatedName = "Pester_Set_Updated_$(Get-Random)"
+        $result = Set-JCWorkflow -Id $script:testWorkflow.id -Name $updatedName -ExecutionRoleId $script:roleId -Dsl $script:dsl
+
+        $result | Should -Not -BeNullOrEmpty
+        $result.id | Should -Be $script:testWorkflow.id
+        $result.name | Should -Be $updatedName
+    }
+
+    It "Throws an error when parameters or ID are invalid" {
+        { Set-JCWorkflow -Id "invalid-id-99999" -Name "Should Fail" } | Should -Throw
```
lucasmendes-jc: The tests expect a runtime error (a throw), so when a 404 or 400 is returned, nothing flags it and the tests still pass; the test needs to expect the specific HTTP request codes.

PowerShell/JumpCloud Module/Public/Workflows/Set-JCWorkflow.ps1
```diff
@@ -0,0 +1,79 @@
...
+        [Parameter(Mandatory = $false, ValueFromPipelineByPropertyName = $true)]
+        [System.String]
+        $Name,
+
+        [Parameter(Mandatory = $false, ValueFromPipelineByPropertyName = $true)]
+        [Alias('WorkflowRole', 'role_id')]
+        [System.String]
+        $ExecutionRoleId,
+
+        [Parameter(Mandatory = $false, ValueFromPipelineByPropertyName = $true)]
```
lucasmendes-jc, line 39: I tested it here and it works perfectly; however, the function lacks abstraction and has quite a few parameters. I suggest setting `DSL` to `Mandatory=$true`, since the documentation lists it as required in the request body.

lucasmendes-jc (changes requested):

PowerShell/JumpCloud Module/Tests/Public/Workflows/Set-JCWorkflow.Tests.ps1
junioralmeida-82: Updated! The negative tests now check for explicit exception messages (`*No update parameters specified*` and `*Failed to update workflow*`) instead of a generic throw, ensuring we validate the exact failure path.

PowerShell/JumpCloud Module/Public/Workflows/Set-JCWorkflow.ps1
junioralmeida-82: Good call out regarding the API docs requiring `dsl` in the payload! However, to support partial updates (e.g., updating only `-Name` or `-Status` without forcing the user to pass the entire DSL JSON again), we implemented a Read-Modify-Write pattern. `Set-JCWorkflow` now fetches the existing workflow via `Get-JCWorkflow` and reuses its `dsl` if omitted in the cmdlet call, so keeping `-Dsl` optional provides a much better PowerShell UX while still fulfilling the API requirements.

junioralmeida-82:
Thanks for the thorough review and testing against the sandbox, @edgar-lins!

I've addressed all the points raised:
1. **Partial updates (Read-Modify-Write):** `Set-JCWorkflow` now retrieves the current workflow via `Get-JCWorkflow` to preserve omitted fields (`dsl`, `execution_role_id`, `description`, `status`) when performing partial updates.
2. **Empty body guard:** Added validation to throw a clear error if no updatable parameters are specified.
3. **Refactored error handling & pipeline:** Enhanced the `catch` block with detailed context and removed `ValueFromPipeline = $true` from `$Id` to prevent unexpected object binding.
4. **Expanded test coverage:** Added test cases for partial updates, empty parameter guard, and specific throw assertions.

All 4 Pester integration tests are now passing locally (screenshot attached below).

lucasmendes-jc (approved):
Sounds good 👍

edgar-lins (approved):
Re-tested against sandbox after the read-modify-write change. Confirmed:

`Set-JCWorkflow -Id X -Name "..."` (name-only) now succeeds and preserves `dsl` — previously returned `602 "dsl in body is required"`. The `dsl` object from `Get-JCWorkflow` round-trips cleanly back through the PUT.
Empty-body guard works and is covered by a test.
The new partial-update test is exactly the regression guard this needed, and the error tests now assert on message.

Two non-blocking notes:

`status` is only included when `$current.status` is non-null — if the API ever requires it in the body (like `dsl`), a null-status workflow could hit the same 602. Worth confirming whether `status` is required on PUT; if so, treat it like the other four fields.
One inline comment is in PT (`# Validar se ao menos um campo editável...`) — rest of the file is EN; worth aligning for consistency.

Neither blocks merge. Nice turnaround on the fix.

